PR #16, opened 2016-08-29
Consolidate similar methods

jnyrup:
Take advantage of the params keyword to get rid of the several similar methods.
Depending on your preference of style you might also like to merge `IsRefactoringEnabled` and `IsAnyRefactoringEnabled`.


JosefPihrt:
Hi, Jonas, thanks for you PR, I appreciate it.

I would stay with existing implementation for performance reasons.

Similar approach can be seen in roslyn internal code here (IsKind extension method):
https://github.com/dotnet/roslyn/blob/master/src/Workspaces/CSharp/Portable/Extensions/SyntaxNodeExtensions.cs

jnyrup:
No worries, it makes perfect sense.
Crazy little thing called ~~love~~ performance.

PR #45, opened 2017-02-18
Fix for RCS1140 adding an extra "/>

NickCraver:
When the `<exception>` is an `ArgumentNullException`, `"/>` was added twice - minor bug. Here's an example of the current behavior before this fix:
screen shot 2017-02-18 at 08 55 50


JosefPihrt:
Thanks for fixing this bug @NickCraver.

PR #112, opened 2017-07-12
Code fixes

jnyrup:
I have found and fixed what I believe to be errors in the code.

Note that I haven't done any testing of it.

### DIFF
--- a/source/Analyzers/CodeFixProviders/ArgumentCodeFixProvider.cs
+++ b/source/Analyzers/CodeFixProviders/ArgumentCodeFixProvider.cs
@@ -9,6 +9,7 @@ using Microsoft.CodeAnalysis.CodeFixes;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Roslynator.CSharp.Refactorings;
 using Roslynator.CSharp.Syntax;
+using System.Diagnostics;
 
 namespace Roslynator.CSharp.CodeFixProviders
 {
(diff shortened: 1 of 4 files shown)


jnyrup:
Damn... I was unsure about that commit. 

I'm not yet familiar enough with the CodeAnalysis framework to know hot to simply fix this.

But the problem is that 
```csharp
if (kind == SyntaxKind.ThisExpression
	|| kind != SyntaxKind.SimpleMemberAccessExpression)
{
	return methodSymbol;
}
```
will pass no matter if `kind == SyntaxKind.ThisExpression` is true or not.

JosefPihrt:
You are right that the condition is incorrect. It should look like this:

```csharp
if (!expression.IsKind(SyntaxKind.SimpleMemberAccessExpression)
    || ((MemberAccessExpressionSyntax)expression).Expression.IsKind(SyntaxKind.ThisExpression))
{
    return methodSymbol;
}
```
(If the invocation expression is qualified then it must be qualified with `this`.)

JosefPihrt:
You are right that the condition is incorrect. It should look like this:

```csharp
if (!expression.IsKind(SyntaxKind.SimpleMemberAccessExpression)
    || ((MemberAccessExpressionSyntax)expression).Expression.IsKind(SyntaxKind.ThisExpression))
{
    return methodSymbol;
}
```
(If the invocation expression is qualified then it must be qualified with `this`.)

JosefPihrt:
Thanks for a contribution!

PR #210, opened 2017-11-26
Bitshifts on long and checks on unsigned

jnyrup:
`(long)(1 << i)` does not work as intended as `1` is implicitly an `int` and `int << i` overflows for `i >= 32`.

The removed checks are trivially true as an unsigned value is always `>= 0`.


JosefPihrt:
Thanks @jnyrup !

PR #362, opened 2018-05-18
Fix typo in RCS1223 docs

xt0rted:
Adds a missing `[` to the code sample

I edited `Analyzers.xml` and then ran the following to regenerate the docs.
```
dotnet .\src\Tools\MetadataGenerator\bin\Debug\netcoreapp2.0\MetadataGenerator.dll src
```

I wasn't sure if anything else needed to be done.


JosefPihrt:
Thanks!

PR #461, opened 2018-11-15
Support global suppression of diagnostics

### LINKED ISSUE #451: Suggestion: Add ability to disable rules without requiring a ruleset
I have recently installed Roslynator and (unsurprisingly) see a lot of noise with the inherited codebase. This codebase does not have a ruleset in the  repository and i'm the only one wib Roslynator installed.  The only way I can find on disabling some of these rules is thru a ruleset. This is not practical in my case since I'm the only one with it installed. Plus, I have to add it to my .gitignore so the ruleset isn't accidentally checked in.

ReSharper has the ability to configure warnings, errors, etc. thru the settings for the extension in VisualStudio. It would be so much more helpful if I could do this.

JosefPihrt:
fixes #451


JosefPihrt:
Current implementation would look like this:

diagnostics

PR #509, opened 2019-02-28
Improve formatting of documentation

Danghor:
- avoids skipping headline levels (### after # without using ##)
- makes table readable in text file
- uses `code` notation for keys


JosefPihrt:
Thanks @Danghor !

PR #524, opened 2019-03-26
Pass cancellation token to methods that support it.

Danghor:
At several places, a cancellation token was being passed to a method, but the method did not pass this cancellation token to other methods that would have supported it. I fixed those place by passing the cancellation token wherever possible.


JosefPihrt:
I'm curious: did you use some analyzer to detect it?

Danghor:
ReSharper... 😅

JosefPihrt:
Thanks!

jnyrup:
@JosefPihrt https://github.com/u2uconsult/codeanalyzers/blob/master/docs/U2U1016.md has a similar feature.

PR #525, opened 2019-03-26
Add more unit tests.

Danghor:
Added some more unit tests to [RCS1227: Validate arguments correctly](https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1227.md) and introduced tests for [RCS1066: Remove empty finally clause](https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1066.md).


Danghor:
Besides the two analyzers mentioned, more unit tests will be pushed to this branch, as I find the time to write them.

JosefPihrt:
Hi, 

I have a couple of notes that might be helpful:

script `tools\TestCodeGenerator.cmd` will generate empty test class for a specified analyzer id(s).
script `tools\test.cmd` will run tests from command line.

Folder `src\Tests\Analyzers.Tests.Old` contains old "manual" tests. The ultimate goal is to transform this code into unit tests in `src\Tests\Analyzers.Tests` and then delete old files.

For example file `src\Tests\Analyzers.Tests.Old\RemoveEmptyFinallyClause.cs` can be deleted when you added unit tests for RCS1066.

Danghor:
Two tests for RCS1043 are currently failing, but it seems like there is an actual bug in the analyzer. When a partial class contains two partial method declarations, no diagnostic is generated for the `partial` modifier in the class:

```csharp
public partial struct Foo // Should generate RCS1043
{
    partial void FooMethod();

    partial void FooMethod()
    {
    }
}
```

JosefPihrt:
I think that this case does not need to be covered by the analyzer because it won't exist in the reality (why have partial definition and implementation in the same file?).

Danghor:
I removed the tests for RCS1007, since they are now included in PR #534.

JosefPihrt:
@Danghor Thanks for the contribution, sorry that it took so long.

PR #532, opened 2019-04-06
Pass CancellationToken on

jnyrup:
Places where a `CancellationToken` is available but not passed on to functions that takes an optional `CancellationToken`.

Some were found by making optional `CancellationToken`s mandatory, others by regex'ing the source code.


Danghor:
Sometimes you use named arguments (`cancellationToken: cancellationToken, `) and sometimes you just pass the parameter (`cancellationToken, `). I think it would be better to use the non-named version to have consistency here.

jnyrup:
I intentionally used named arguments whenever necessary to avoid specifying additional optional arguments.
Searching the existing codebase there it at least 22 places where there is a mix of named and non-named arguments.

Danghor:
Oh, that makes sense. I didn't notice they were necessary to avoid specifying optional arguments.

jnyrup:
@JosefPihrt Can you comment on whether this PR is useful or should just be closed?
If it is useful, is there anything I need to do, besides resolving the conflict, before it is ready?

JosefPihrt:
I will give you feedback shortly.

PR #534, opened 2019-04-10
Prevent reformatting when adding braces via codefix

Danghor:
While writing the unit tests for #525, I noticed some tests breaking because the `AddBracesRefactoring` did not correctly preserve the trivia and changed the formatting of the code. After the change, only braces are inserted, the trivia is preserved and the formatting of the code remains unchanged.


JosefPihrt:
I don't understand why statement's trivia should be moved outside the block. Could you add some unit that will clarify the intent of this PR.

And if the trivia should be moved then it is necessary to remove it from the statement, otherwise it will appear twice.

Danghor:
Maybe my proposed solution was wrong.
I tested it using the Unitests in `src/Tests/Analyzers.Tests/RCS1007AddBracesTests.cs` in the branch [Danghor:unitTests](https://github.com/Danghor/Roslynator/tree/unitTests) that is used in pull request #525. Should these tests be included in this PR, rather than the other one?

JosefPihrt:
Yes, please include them in this PR.

Danghor:
It seems like the issue is not related to the Trivia, but to the `WithFormatterAnnotation();` statement at the end of the line.

I tried out the current version of Roslynator (2017 2.1.0 installed in Visual Studio 2017) with the following example:

Before

On this example, nothing changes when I reformat the code manually via  `Ctrl` + `K`, `Ctrl` + `D`.

When I use the Roslynator Code Fix to add braces, I get the following result:

After

Again, reformatting manually does not change the code. However, the two lines with the arguments `x` and  `y` move to the left one space, which also breaks the unit tests attached. When I use the built-in refactoring from Visual Studio (IDE0011), the result is as follows:

Expected

This code is also unaffected by a manual reformatting.

When the `WithFormatterAnnotation();` statement is ommitted, the unit tests pass, regardless of the trivia statements. I will therefore add a commit which removes those trivia statements since they do not seem correct here.

JosefPihrt:
Looks good, thanks!

PR #536, opened 2019-04-15
Remove null-check for the body-block when analyzing a catch clause (Issue #527)

### LINKED ISSUE #527: Null check for catch clause
In the `RemoveOriginalExceptionFromThrowStatementAnalyzer`, there is a null check for the catch clause. I am currently struggeling with covering this line with a unit test. How could the catch block ever be `null`? Even if it is empty, it would not be `null` then.

https://github.com/JosefPihrt/Roslynator/blob/fecd10812f7c48869d6720a031e40d08cb9d374d/src/Analyzers/CSharp/Analysis/RemoveOriginalExceptionFromThrowStatementAnalyzer.cs#L31-L40

Danghor:
Removes the null check for the block in a catch clause, since it cannot be null


JosefPihrt:
Please change title to something more descriptive.

JosefPihrt:
fixes #527

PR #545, opened 2019-05-06
PrefixFieldIdentifierWithUnderscore  default

davidroth:
Set PrefixFieldIdentifierWithUnderscore default to `false` because this is the visual studio default.

**Rational:**
I know that there are two camps regarding the underscore prefix and I don`t wanna start arguing whether one is better than the other.

However, I think it makes more sense that roslynator respects the visual studio **default** in this case and not modifies the default.
Roslynator adds a ton of value by adding awesome refactorings and fixes and I think it is not the users expectation that it modifies its default CSharp code style. (At least it was not mine)

If one really likes the underscore prefix it is super easy to enable it, but I believe that roslynator should not change the default style in this regards.

### DIFF
--- a/src/Common/RefactoringSettings.cs
+++ b/src/Common/RefactoringSettings.cs
@@ -6,11 +6,11 @@ namespace Roslynator
     {
         public static RefactoringSettings Current { get; } = new RefactoringSettings();
 
-        public bool PrefixFieldIdentifierWithUnderscore { get; set; } = true;
+        public bool PrefixFieldIdentifierWithUnderscore { get; set; }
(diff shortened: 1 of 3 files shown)


JosefPihrt:
Hi,

Sorry for the delay. I'm fine with this proposal I would just introduce it into Roslynator 2019 only.

Could you change base branch to 2019?

davidroth:
> Could you change base branch to 2019?

Sure, I have changed the target branch and rebased.

JosefPihrt:
Thanks.

PR #564, opened 2019-06-24
Add refactorings to inline all instances of a method.

ILMTitan:


ILMTitan:
This implementation is not good enough. May try again later.

PR #575, opened 2019-07-24
Analyzers/use string empty instead of empty string

amakhno:
Hi!

I'm sorry for probably bad code, but it's my first open-source contribution experience. 
Will be glad to know what's wrong with my code (I know that it is).
Your wiki is a bit poor, and it was difficult to find out not the best solution of this refactoring, but working one.

Thank you,
Artem


src/Analyzers.CodeFixes/CSharp/CodeFixes/UseStringEmptyInsteadOfEmptyStringLiteralCodeFixProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+            if (!TryFindFirstAncestorOrSelf(root, context.Span, out SyntaxNode node, predicate: f => f.IsKind(
+                SyntaxKind.StringLiteralExpression)))
+            {
+                return;
+            }
+
+            Diagnostic diagnostic = context.Diagnostics[0];
+
+            var codeAction = CodeAction.Create(
+                "Use StringюEmpty instead of \"\"",
```
De-Su, line 31: ю)

JosefPihrt:
https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1078

PR #636, opened 2020-01-28
Add closing single quote

camleng:


JosefPihrt:
Hi, these files are generated and cannot be edited directly. Please follow instruction in this file:
https://github.com/JosefPihrt/Roslynator/blob/master/docs/HowToUpdateDocumentation.md

PR #648, opened 2020-02-20
Add open configuration commands to Command Palette (VSCode)

AdrianWilczynski:
Hi, this contributes Command Palette entries for opening Roslynator's configuration files: `Roslynator: Open Configuration of Analyzers (roslynator.ruleset)` and `Roslynator: Open Configuration of Refactorings and Fixes (roslynator.config)`.

They create (if necessary - non-existent) configuration files and open them. In practice, they work like built-in `Preferences: Open Settings (JSON)` command.

You can call them by pressing `Ctrl + Shift + P` and typing their name.

This PR also moves omnisharp.json update out of the `extension.ts` file and bumps required VSCode version to be safe (I think I remember that recursive directory creation didn't work in VSCode a year ago, It works now and I'm just not sure when It was added; this PR can be made to work without this change though).

You can find tests for this functionality in `openRoslynatorConfiguration.test.ts`.

Let me know if this is something you'd be open to adding and if I should change/fix something before.


JosefPihrt:
Overally it looks fine.

Could you just change `configurationFiles.ts` to `configurationFiles.generated.ts`. In a future, I would like to create default content of ruleset and config file from one place.

AdrianWilczynski:
Sure thing. I've changed it.

Issue #664, opened 2020-04-21
Update omnisharp.json when uninstalling Roslynator


JosefPihrt:
Currently, references to Roslynator libraries are not removed from omnisharp.json when Roslynator extension is uninstalled.

Because uninstalled extensions are not removed from `.vscode/extensions`, Roslynator libraries are loaded even if extension has been uninstalled.

@AdrianWilczynski Would you be interested to do a PR? I think it would be a piece of cake for you.

AdrianWilczynski:
Yeah, sure. I see there is an uninstall hook (https://code.visualstudio.com/api/references/extension-manifest#extension-uninstall-hook) and I can probably clean up `omnisharp.json`using it. I'll try it.

(Extensions do seem to get removed from `extensions` directory after uninstall but it seems to happen only after restarting VSCode. I guess even if that happens references pointing to not existing libraries might still be causing some trouble.)

JosefPihrt:
Thanks!

JosefPihrt:
Fixed with #666

PR #666, opened 2020-04-21
Cleanup on uninstall

AdrianWilczynski:
This should work for #664.

Uninstall hook seems to only run after full restart so you have to close and reopen VSCode, just clicking "Reload Required" button after uninstalling doesn't seem to be enough.

Let me know if there is something I should change or if you can spot any bugs.


JosefPihrt:
Fixes #664

JosefPihrt:
Thanks!

PR #699, opened 2020-07-26
Update RCS1166.md

xtqqczze:
Add additional example for a non-numeric type in documentation for `RCS1166: Value type object is never equal to null`.

### DIFF
--- a/docs/analyzers/RCS1166.md
+++ b/docs/analyzers/RCS1166.md
@@ -12,10 +12,15 @@
 
 ```csharp
 int x = 0;
+IntPtr y = IntPtr.Zero
 
 // ...
 
 if (x == null) // RCS1166
+{
+}
+
+if (y == null) // RCS1166
 {
 }
 ```
@@ -24,6 +29,10 @@ if (x == null) // RCS1166
 
 ```csharp
 if (x == 0)
+{
+}
+
+if (y == default)
 {
 }
 ```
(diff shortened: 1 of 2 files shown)


JosefPihrt:
The current fix replaces `null` with `0` not `default`. Am I missing something?

xtqqczze:
> 
> 
> The current fix replaces `null` with `0` not `default`. Am I missing something?

https://github.com/xtqqczze/PowerShell/commit/4c33ee0dee4539c56a03c42a45b7141ed6316891

xtqqczze:
@JosefPihrt My bad, I missed in your example `x` is type `int`, so `0` would be correct.

xtqqczze:
I think the example is more helpful with a non-numeric type like `IntPtr` where it is less obvious it is never equal to null.

JosefPihrt:
I'm fine with adding another example, I don't think it is necessary to overwrite original example.

xtqqczze:
@JosefPihrt New example added, my branch is open if you want to make any changes.

JosefPihrt:
I forgot to tell you that the file you edited is actually generated.

Please follow [these](https://github.com/JosefPihrt/Roslynator/blob/master/docs/HowToUpdateDocumentation.md) instructions.

xtqqczze:
@JosefPihrt updated

PR #745, opened 2020-11-12
Fix CS0177

xtqqczze:
https://docs.microsoft.com/dotnet/csharp/misc/cs0177

I attempted to build commandline project using new net5.0 tfm, but the build failed with CS0177: The out parameter 'parameter' must be assigned to before control leaves the current method.

P.S. Do you have plans to support .NET 5.0?


JosefPihrt (changes requested):
I would rather return "default" instead of empty array.

JosefPihrt:
Are there any benefits from moving CLI to .NET 5.0? I mean, it will eventually move to .NET 5.0 but there is no need to hurry.

JosefPihrt (approved):

JosefPihrt:
Thanks!

Issue #538, opened 2019-04-18
Suggestion: Use Keys/Values enumerators when possible, instead of KeyValuePair enumerator


wiz0u:
Collections of type `IDictionary<TKey, TValue>` implement `IEnumerable<KeyValuePair<TKey, TValue>>` and also offer direct `Values` and `Keys` enumerators.

For these collections, I would like Roslynator to identify this kind of LINQ expression:
`collection.First().Key` or `collection.First().Value`
and suggest to refactor it as
`collection.Keys.First()` or `collection.Values.First()`

I feel the `Keys` and `Values` enumerators are always more efficient than the construction of several `KeyValuePair` (not usually a native construction of the collection) just for the purpose of accessing their `.Key` or `.Value` component

These suggestions should at least be compatible with these LINQ methods: `ElementAt`, `ElementAtOrDefault`, `First()`, `FirstOrDefault()`, `Last()`, `LastOrDefault()`, `Single()`, `SingleOrDefault()`.

To go further, `Min`, `Max`, `Sum`, `Average`, `Count` with selectors, and overloads of the preceding methods taking a predicate, could be also be compatible if it is identified that only the adequate `.Key` or `.Value` component is used on the lambda-expression argument `KeyValuePair`

josefpihrt:
I know that this is a very old issue, but let me express my opinion:

1) I think that calling `dictionary.First()` (or similar LINQ method) on a dictionary is very questionable. Why would you want to obtain first item in a dictionary when you know that items in a dictionary are sorted in an undefined order? Maybe you could call it on a sorted dictionary, but still it would not be more efficient:

- For `dictionary.First().Value` it's necessary to allocate enumerator for IEnumerable and KeyValuePair is a struct.
- For `dictionary.Values.First()` it's necessary to initialize collection of values + allocate enumerator for IEnumerable.

https://github.com/dotnet/runtime/blob/6fafbad67235cc589ff3e1928e764e22e7bd1d80/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L222

PR #766, opened 2021-01-26
Use Keys/Values enumerators when possible, instead of KeyValuePair enumerator

TryNewJais:
This Pull request resolves  the Issues #538.


josefpihrt:
See https://github.com/dotnet/roslynator/issues/538#issuecomment-1902624733

PR #928, opened 2022-07-10
Code fix for RCS1197 should add parentheses if necessary

### LINKED ISSUE #923: RCS1197 bug in optimization suggest when calculation is involved.
**Product and Version Used**: 

Roslynator 4.11

**Steps to Reproduce**:

Given this code

```csharp
sb.Append(s.Substring(start + len));
```

**Actual Behavior**:

The suggested optimization is

```csharp
sb.Append(s, start + len, s.Length - start + len);
```

but possibly should be

**Expected Behavior**:
```csharp
sb.Append(s, start + len, s.Length - (start + len));
```

karl-sjogren:
Fixes #923.

This only takes care of the example in the issue, it doesn't consider if there are other cases where this should be done as well.


src/Analyzers.CodeFixes/CSharp/CodeFixes/OptimizeStringBuilderAppendCallCodeFixProvider.cs
```diff
@@ -258,13 +258,19 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+                                    }
+
                                     ArgumentListSyntax argumentList = ArgumentList(
                                         Argument(invocationInfo.Expression),
                                         arguments[0],
                                         Argument(
                                             SubtractExpression(
                                                 SimpleMemberAccessExpression(invocationInfo.Expression, IdentifierName("Length")),
-                                                arguments[0].Expression)));
+                                                argumentExpression)));
```
JosefPihrt, line 273: Instead of ParenthesizedExpression use `arguments[0].Expression.Parenthesize()`. The parentheses will be removed if not necessary.

JosefPihrt (changes requested):

src/Analyzers.CodeFixes/CSharp/CodeFixes/OptimizeStringBuilderAppendCallCodeFixProvider.cs
karl-sjogren: Wow, there are so many great helper methods in this project! Updated the code.

JosefPihrt (approved):

JosefPihrt:
Thanks!

PR #1031, opened 2023-01-24
Fix CS0164 CodeFixProvider

jamesHargreaves12:
Previous behaviour would delete the child statement of the label. The correct behaviour is just to remove the label.

### DIFF
--- a/src/CodeFixes/CSharp/CodeFixes/LabeledStatementCodeFixProvider.cs
+++ b/src/CodeFixes/CSharp/CodeFixes/LabeledStatementCodeFixProvider.cs
@@ -2,8 +2,10 @@
 
 using System.Collections.Immutable;
 using System.Composition;
+using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


src/CodeFixes/CSharp/CodeFixes/LabeledStatementCodeFixProvider.cs
```diff
@@ -30,7 +32,15 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
 
         if (!TryFindFirstAncestorOrSelf(root, context.Span, out LabeledStatementSyntax labeledStatement))
             return;
+        
+        var child = labeledStatement.ChildNodes().First();
```
JosefPihrt, line 36: Why not use just `labeledStatement.Statement`?

JosefPihrt (changes requested):

JosefPihrt (approved):

PR #1037, opened 2023-02-13
Fix for RemoveRedundantToStringCall(RCS1097) 

jamesHargreaves12:
Current behaviour:
- RCS1097 suggest removing ToString for structs which can lead to boxing.

It looks as if this was attempted to be prevented by using an IsReferenceType check at:
https://github.com/JosefPihrt/Roslynator/blob/3c6f1c1fc24b63aea6b0d82e64ec4e02d9ca39ad/src/Analyzers/CSharp/Analysis/RemoveRedundantToStringCallAnalysis.cs#L60

Unfortunately, this doesn't work for structs as they have type System.ValueType which is considered a reference type. See
https://learn.microsoft.com/en-us/dotnet/api/system.type.isvaluetype?redirectedfrom=MSDN&view=net-7.0#System_Type_IsValueType

Fix implemented just explicitly checks for `System_ValueType`.

Also added a test to catch future regressions.

### DIFF
--- a/src/Tests/Analyzers.Tests/RCS1097RemoveRedundantToStringCallTests.cs
+++ b/src/Tests/Analyzers.Tests/RCS1097RemoveRedundantToStringCallTests.cs
@@ -105,7 +105,7 @@ class C
     }
 
     [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.RemoveRedundantToStringCall)]
-    public async Task TestNoDiagnostic_int()
+    public async Task TestNoDiagnostic_Int()
     {
         await VerifyNoDiagnosticAsync(@"
 class C
--- a/src/Tests/Analyzers.Tests/RCS1097RemoveRedundantToStringCallTests.cs
+++ b/src/Tests/Analyzers.Tests/RCS1097RemoveRedundantToStringCallTests.cs
@@ -105,7 +105,7 @@ class C
     }
 
     [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.RemoveRedundantToStringCall)]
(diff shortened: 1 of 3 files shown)


src/Tests/Analyzers.Tests/RCS1097RemoveRedundantToStringCallTests.cs
```diff
@@ -105,7 +105,7 @@ void M()
     }
 
     [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.RemoveRedundantToStringCall)]
-    public async Task TestNoDiagnostic_ValueType()
+    public async Task TestNoDiagnostic_int()
```
JosefPihrt, line 108:
```suggestion
    public async Task TestNoDiagnostic_Int()
```

src/Tests/Analyzers.Tests/RCS1097RemoveRedundantToStringCallTests.cs
```diff
@@ -116,6 +116,22 @@ void M()
         string s = $""'{i.ToString()}'"";
     }
 }
+");
+    }
+    
+    [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.RemoveRedundantToStringCall)]
+    public async Task TestNoDiagnostic_struct()
```
JosefPihrt, line 123:
```suggestion
    public async Task TestNoDiagnostic_Struct()
```

JosefPihrt (changes requested):

JosefPihrt:
Hi,

thanks for the PR, I just made some minor comments.

PR #1039, opened 2023-02-15
Stop RCS1031, RCS1208 and RCS1211 from being applied if the local variables overlap.

jamesHargreaves12:
For cases, where the same local variable is declared inside the if and else blocks e.g.
```
if (flag)
{
    var x  = 1;
    return x;
}
else 
{
    var x = 0;
    return x;
}
```
The diagnostic RCS1211 Should not be raised. 

Removing the else block would cause a compiler error "A local variable named 'x' cannot be declared in this scope because it would give a different meaning to 'x', which is already used in a parent or current scope to denote something else".

The same thing is true for switch case brackets (RCS1031).

### DIFF
--- a/src/Analyzers/CSharp/Analysis/RemoveUnnecessaryBracesInSwitchSectionAnalyzer.cs
+++ b/src/Analyzers/CSharp/Analysis/RemoveUnnecessaryBracesInSwitchSectionAnalyzer.cs
@@ -107,22 +107,29 @@ public sealed class RemoveUnnecessaryBracesInSwitchSectionAnalyzer : BaseDiagnos
     
     private static bool LocalDeclaredVariablesOverlapWithAnyOtherSwitchSections(SwitchStatementSyntax switchStatement, BlockSyntax caseBlock, SemanticModel semanticModel)
     {
-        var caseDeclaredLocalVars = semanticModel
-            .LookupSymbols(caseBlock.CloseBraceToken.SpanStart-1)
-            .Where(s => s.Kind == SymbolKind.Local)
-            .Where(s => s.Locations.All(l => caseBlock.Span.Contains(l.SourceSpan)))
+        var caseVariablesDeclared = semanticModel.AnalyzeDataFlow(caseBlock)?
+            .VariablesDeclared;
+        
+        if (!caseVariablesDeclared.HasValue || caseVariablesDeclared.Value == ImmutableArray<ISymbol>.Empty)
+            return false;
+
+        var nameCaseDeclaredVariables = caseVariablesDeclared.Value
             .Select(s => s.Name)
             .ToImmutableHashSet();
 
+        
         // Now check if any of the other case statements under the same switch contain definitions for the same local variable. 
         return switchStatement.Sections.Any(section =>
         {
             if (section.Statements.SingleOrDefault(shouldThrow: false) is not BlockSyntax block || block == caseBlock)
                 return false;
 
-            return semanticModel
-                .LookupSymbols(block.CloseBraceToken.SpanStart-1)
-                .Any(s => s.Kind == SymbolKind.Local && caseDeclaredLocalVars.Contains(s.Name));
+            var sectionVariablesDeclared = semanticModel.AnalyzeDataFlow(block)?.VariablesDeclared;
+            if (sectionVariablesDeclared is null || sectionVariablesDeclared.Value.IsEmpty)
+                return false;
+
+            return sectionVariablesDeclared.Value.Any(s => nameCaseDeclaredVariables.Contains(s.Name));
+
         });
     }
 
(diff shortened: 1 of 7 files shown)


src/Analyzers/CSharp/Analysis/RemoveUnnecessaryBracesInSwitchSectionAnalyzer.cs
```diff
@@ -98,4 +104,27 @@ static bool AnalyzeTrivia(SyntaxTriviaList trivia)
...
+    
+    private static bool LocalDeclaredVariablesOverlapWithAnyOtherSwitchSections(SwitchStatementSyntax switchStatement, SwitchSectionSyntax switchSection, SemanticModel semanticModel)
+    {
+        var sectionVariablesDeclared = semanticModel.AnalyzeDataFlow(switchSection)!
+            .VariablesDeclared;
+        
+        if (sectionVariablesDeclared.IsEmpty)
+            return false;
+
+        var sectionDeclaredVariablesName = sectionVariablesDeclared
```
JosefPihrt, line 116: `Name` should be `Names` so it's clear that it is a collection.

src/Analyzers/CSharp/Analysis/RemoveUnnecessaryBracesInSwitchSectionAnalyzer.cs
```diff
@@ -98,4 +104,27 @@ static bool AnalyzeTrivia(SyntaxTriviaList trivia)
...
+            .ToImmutableHashSet();
+
+        return switchStatement.Sections.Any(otherSection =>
+        {
+            var otherSectionVariablesDeclared = semanticModel.AnalyzeDataFlow(otherSection)!.VariablesDeclared;
+            if (otherSectionVariablesDeclared.IsEmpty)
+                return false;
+
+            return otherSectionVariablesDeclared.Any(s => sectionDeclaredVariablesName.Contains(s.Name));
+        });
```
JosefPihrt, line 127: I would prefer to use nested foreach loop to get rid of allocations.

Something like:
```
foreach otherSection in switchStatement.Sections
{
    foreach v in semanticModel.AnalyzeDataFlow(otherSection)!.VariablesDeclared
    {
        if (sectionDeclaredVariablesName.Contains(v.Name))
            return true;
    }
}

return false;
```

+ you should skip the section that is being analyzed.

src/Analyzers/CSharp/Analysis/RemoveUnnecessaryElseAnalyzer.cs
```diff
@@ -55,12 +57,39 @@ public static bool IsFixable(ElseClauseSyntax elseClause)
...
+            return false;
 
-        return statement is not null
-            && CSharpFacts.IsJumpStatement(statement.Kind());
+        var elseVariableNames = elseVariablesDeclared
+            .Select(s => s.Name)
+            .ToImmutableHashSet();
+        
+        return ifVariablesDeclared
+            .Any(s => elseVariableNames.Contains(s.Name));
```
JosefPihrt, line 93: Same as comment above: better would be to use foreach loop to avoid allocations.

JosefPihrt (changes requested):

src/Analyzers/CSharp/Analysis/RemoveUnnecessaryBracesInSwitchSectionAnalyzer.cs
jamesHargreaves12: Skipping the section being analyzed is a good point. I have added a test to catch this.

src/Common/CSharp/Analysis/ReduceIfNesting/ReduceIfNestingAnalysis.cs
```diff
@@ -71,6 +72,11 @@ private static ReduceIfNestingAnalysisResult Fail(SyntaxNode topNode)
                 return Fail(switchSection);
             }
 
+            if (LocallyDeclaredVariablesOverlapWithOuterScope(ifStatement, switchSection, semanticModel))
+            {
+                return Fail(switchSection);
+            }
```
JosefPihrt, line 78:
```suggestion
            if (LocallyDeclaredVariablesOverlapWithOuterScope(ifStatement, switchSection, semanticModel))
                return Fail(switchSection);
```

src/Common/CSharp/Analysis/ReduceIfNesting/ReduceIfNestingAnalysis.cs
```diff
@@ -104,6 +110,11 @@ private static ReduceIfNestingAnalysisResult Fail(SyntaxNode topNode)
                 return Fail(parent);
             }
 
+            if (LocallyDeclaredVariablesOverlapWithOuterScope(ifStatement, parent, semanticModel))
+            {
+                return Fail(parent);
+            }
```
JosefPihrt, line 116:
```suggestion
            if (LocallyDeclaredVariablesOverlapWithOuterScope(ifStatement, parent, semanticModel))
                return Fail(parent);
```

JosefPihrt (approved):

JosefPihrt:
Thanks for the PR! 👍

PR #1051, opened 2023-03-13
Fix handling of records in RCS1223

jamesHargreaves12:
Currently, this analyzer will throw the following error when it comes across a record:
`System.Exception : Analyzer execution threw exception Analyzer 'Roslynator.CSharp.Analysis.MarkTypeWithDebuggerDisplayAttributeAnalyzer' threw an exception of type 'System.InvalidCastException' with message 'Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Syntax.RecordDeclarationSyntax' to type 'Microsoft.CodeAnalysis.CSharp.Syntax.ClassDeclarationSyntax'.'.
`
This shows up as an AD0001 Diagnostic.

Casting Instead to TypeDeclerationSyntax means that we can also handle records correctly. There is no behaviour change for either Struct or Class if we cast to TypeDeclerationSyntax.


JosefPihrt:
Nice! 👍

PR #1058, opened 2023-03-21
Fix/RCS1031 and RCS1208

jamesHargreaves12:
This PR extends the local variable check from https://github.com/JosefPihrt/Roslynator/pull/1039 to also check variables assigned during pattern matching for RCS1031.

Slightly embarrassingly, I didn't check all the ReduceIfNestingAnalysis code paths in https://github.com/JosefPihrt/Roslynator/pull/1039 and so I didn't realise that the AnalyzeDataFlow can only handle an argument of type ExpressionSyntax or StatementSyntax. If you pass it anything else then it will throw a run time error. This PR fixes this issue and adds a number of new tests to enforce.

It is currently not trivial to test the Loop / Switch sections of ReduceIfNestingAnalysis.AnalyzeCore via the analyzer as it passes through `ReduceIfNestingOptions.None`. However, I manually removed the guard clause on lines 63 and 90 and confirmed that it was working correctly.


JosefPihrt:
@jamesHargreaves12 What was the reason for closing this issue?

jamesHargreaves12:
> @jamesHargreaves12 What was the reason for closing this PR?
Yesterday while testing something else I found another change that made sense to be included within this PR but did not have time to add unit tests and create a PR for it. So I closed the PR to avoid you spending any time reviewing it until I had a chance to include those changes. The PR with the updates is here https://github.com/JosefPihrt/Roslynator/pull/1062

JosefPihrt:
@jamesHargreaves12 👍

PR #1062, opened 2023-04-03
Fix/RCS1031, RCS1208 and RCS1061

jamesHargreaves12:
This PR extends the local variable check from https://github.com/JosefPihrt/Roslynator/pull/1039 to RCS1061 and also checks variables assigned during pattern matching in the case of RCS1031.

Slightly embarrassingly, I didn't check all the ReduceIfNestingAnalysis code paths in https://github.com/JosefPihrt/Roslynator/pull/1039 and so I didn't realise that the AnalyzeDataFlow can only handle an argument of type ExpressionSyntax or StatementSyntax. If you pass anything else, it will throw a run time error. This PR solves this issue through the new helper function `IfStatementLocalVariableAnalysis.TryGetOuterScope` which handles the intricacies of this. I have also added a number of new tests at the level of RCS1208 to enforce this behaviour.

It is currently not trivial to test the Loop / Switch sections of ReduceIfNestingAnalysis.AnalyzeCore via the analyzer as it passes through ReduceIfNestingOptions.None. However, I manually removed the guard clause on lines 63 and 90 and confirmed that it was working correctly.


JosefPihrt (changes requested):

JosefPihrt:
@jamesHargreaves12 I noticed that there multiple small fixes like formatting, trailing whitespace, implicit/explicit type etc. I suggest to run script `tools/cli_fix.ps1` to fix those.

JosefPihrt:
@jamesHargreaves12 I would like to release new version in about one week and I would like to include this PR in the new version. Could you let me know if you will have time to address my comments so the PR can me merged? Thanks.

src/Common/CSharp/Analysis/ReduceIfNesting/IfLocalVariableAnalysis.cs
```diff
@@ -0,0 +1,73 @@
...
+        
+        var outerScope = GetOuterScope(ifStatement);
+        
+        if (outerScope == null)
+            return true;
+
+        IList<ISymbol> parentVariablesDeclared;
+        if (outerScope is SwitchSectionSyntax switchSection)
+        {
+            parentVariablesDeclared = switchSection.Statements.SelectMany(s => semanticModel.AnalyzeDataFlow(s)!.VariablesDeclared).ToList();
```
JosefPihrt, line 34: By lazy initialization I meant following:

```
List<ISymbol> allDeclaredVariables = null;
foreach (StatementSyntax statement in switchSection.Statements)
{
    ImmutableArray<ISymbol> variables = semanticModel.AnalyzeDataFlow(statement)!.VariablesDeclared;

    if (variables.Any())
        (allDeclaredVariables ??= new List<ISymbol>()).AddRange(variables);
}

parentVariablesDeclared = allDeclaredVariables;
```

src/CSharp/PattenMatchingVariableDeclarationHelper.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Roslynator.CSharp.Analysis;
+
+public static class PattenMatchingVariableDeclarationHelper
+{
+    public static IEnumerable<string> GetVariablesDeclared(PatternSyntax patternSyntax)
```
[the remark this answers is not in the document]
JosefPihrt: fixed

JosefPihrt (changes requested):

src/Common/CSharp/Analysis/ReduceIfNesting/IfLocalVariableAnalysis.cs
JosefPihrt: I took a look at this part of the code and I think that parent scope will be always block or switch section. If the parent scope is, let's say, foreach statement then it means that the if statement is actually an embedded statement (and current implementation will return if statement itself as a parent scope). And if statement's parent will never be method, constructor etc.

So I propose following changes that would simplify the code:

(`StatementListInfo` is basically something that contains a list of statements, i.e. block or switch section)

```cs
foreach (StatementSyntax statement in SyntaxInfo.StatementListInfo(ifStatement).Statements)
{
    if (statement == ifStatement)
        continue;

    foreach (ISymbol parentVariable in semanticModel.AnalyzeDataFlow(statement)!.VariablesDeclared)
    {
        foreach (ISymbol ifVariable in ifVariablesDeclared)
        {
            if (ifVariable.Name == parentVariable.Name)
                return true;
        }
    }
}
```

jamesHargreaves12: Yes I think that makes sense.

JosefPihrt (approved):

JosefPihrt (approved):

PR #1063, opened 2023-04-03
Fix RCS1009's handling of discard designations.

jamesHargreaves12:
Currently, if you try to apply RCS1009 on a for each which contains discards e.g.
```
foreach(var (x,_) in items){...}
```
Then it will fail since these do not get parsed as DeclarationExpressions at Document Refactorings [L92](https://github.com/JosefPihrt/Roslynator/blob/78e385f810812efb6aec7b47ecf7740d80870cd6/src/Workspaces.Common/CSharp/DocumentRefactorings.cs#L92). This PR fixes this by adding cases for PredefinedTypeSyntax and MemberAccessExpression which are what `string _` and `System.DateTime _` get parsed to respectively.

### DIFF
--- a/src/Workspaces.Common/CSharp/DocumentRefactorings.cs
+++ b/src/Workspaces.Common/CSharp/DocumentRefactorings.cs
@@ -97,14 +97,11 @@ internal static class DocumentRefactorings
             {
                 if (f.Expression is DeclarationExpressionSyntax declarationExpression)
                     return f.WithExpression(declarationExpression.WithType(declarationExpression.Type.WithSimplifierAnnotation()));
-                if (f.Expression is PredefinedTypeSyntax predefinedTypeSyntax)
-                    return f.WithExpression(DeclarationExpression(predefinedTypeSyntax, DiscardDesignation()));
-                if (f.Expression is MemberAccessExpressionSyntax memberAccessExpressionSyntax)
+                if (f.Expression is PredefinedTypeSyntax or MemberAccessExpressionSyntax)
                 {
-                    TypeSyntax typeSyntax = GetQualifiedName(memberAccessExpressionSyntax).WithSimplifierAnnotation();
-                    return f.WithExpression(DeclarationExpression(typeSyntax, DiscardDesignation()));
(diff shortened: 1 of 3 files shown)


src/Workspaces.Common/CSharp/DocumentRefactorings.cs
```diff
@@ -97,6 +97,13 @@ private static TupleExpressionSyntax CreateTupleExpression(ITypeSymbol typeSymbo
             {
                 if (f.Expression is DeclarationExpressionSyntax declarationExpression)
                     return f.WithExpression(declarationExpression.WithType(declarationExpression.Type.WithSimplifierAnnotation()));
+                if (f.Expression is PredefinedTypeSyntax predefinedTypeSyntax)
+                    return f.WithExpression(DeclarationExpression(predefinedTypeSyntax, DiscardDesignation()));
+                if (f.Expression is MemberAccessExpressionSyntax memberAccessExpressionSyntax)
+                {
+                    TypeSyntax typeSyntax = GetQualifiedName(memberAccessExpressionSyntax).WithSimplifierAnnotation();
+                    return f.WithExpression(DeclarationExpression(typeSyntax, DiscardDesignation()));
+                }
```
JosefPihrt, line 106: I'd use more simple:

```cs
if (f.Expression is PredefinedTypeSyntax or MemberAccessExpressionSyntax)
{
    return f.WithExpression(DeclarationExpression(ParseTypeName(f.Expression.ToString()).WithSimplifierAnnotation(), DiscardDesignation()));
}
```

and remove method `GetQualifiedName`.

JosefPihrt (changes requested):

src/Workspaces.Common/CSharp/DocumentRefactorings.cs
```diff
@@ -106,7 +110,7 @@ private static TupleExpressionSyntax CreateTupleExpression(ITypeSymbol typeSymbo
 
         return tupleExpression.WithArguments(newArguments);
     }
-
+
```
JosefPihrt, line 113:
```suggestion

```

PR #1066, opened 2023-04-05
File specific configuration should be respected over global configuration.

jamesHargreaves12:
With config from .globalconfig, roslynator analyzers should respect file-specific diagnostic severities over those set globally.

For example, if we have a .globalconfig file that looks like the following:
```
is_global = true
dotnet_diagnostic.RCS1007.severity = none

[/dev/repo/File.cs]
dotnet_diagnostic.RCS1007.severity = warning
```
The current behaviour is that RCS1007 will not be applied to any files.

I would expect the diagnostic to be applied to File.cs but not the rest of the files in the repository (which is consistent with how analyzers from Roslyn (and the majority of roslynator analyzers) work.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Do not remove braces in the cases where there are overlapping local variables. ([RCS1031](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1013.md), [RCS1211](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1211.md), [RCS1208](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1208.md)) ([#1039](https://github.com/JosefPihrt/Roslynator/pull/1039)).
 - [CLI] Analyze command does not create the XML output file and returns incorrect exit code when only compiler diagnostics are reported ([#1056](https://github.com/JosefPihrt/Roslynator/pull/1056) by @PeterKaszab).
 - [CLI] Fix exit code when multiple projects are processed ([#1061](https://github.com/JosefPihrt/Roslynator/pull/1061) by @PeterKaszab).
+- Fix code fix for CS0164 ([#1031](https://github.com/JosefPihrt/Roslynator/pull/1031)).
+
 
 ## [4.2.0] - 2022-11-27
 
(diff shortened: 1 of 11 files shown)


src/Core/Extensions/DiagnosticsExtensions.cs
```diff
@@ -556,8 +556,8 @@ internal static bool IsEffective(this DiagnosticDescriptor descriptor, SyntaxNod
         SyntaxTreeOptionsProvider provider = compilationOptions.SyntaxTreeOptionsProvider;
 
         if (provider is not null
-            && !provider.TryGetGlobalDiagnosticValue(descriptor.Id, cancellationToken, out reportDiagnostic)
-            && !provider.TryGetDiagnosticValue(syntaxTree, descriptor.Id, cancellationToken, out reportDiagnostic))
+            && !provider.TryGetDiagnosticValue(syntaxTree, descriptor.Id, cancellationToken, out reportDiagnostic)
+            && !provider.TryGetGlobalDiagnosticValue(descriptor.Id, cancellationToken, out reportDiagnostic))
```
JosefPihrt, line 560: You are right that the order is incorrect.

I think that the correct order should be file/project/global and "project-wide" configuration is stored in `compilationOptions.SpecificDiagnosticOptions`:

```
if (provider?.TryGetDiagnosticValue(syntaxTree, descriptor.Id, cancellationToken, out reportDiagnostic) != true
    && !compilationOptions.SpecificDiagnosticOptions.TryGetValue(descriptor.Id, out reportDiagnostic))
{
    provider?.TryGetGlobalDiagnosticValue(descriptor.Id, cancellationToken, out reportDiagnostic);
}
```

Also please apply similar order in the method `GetEffectiveSeverity` which is located several lines below.

JosefPihrt (changes requested):

ChangeLog.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix code fix for CS0164 ([#1031](https://github.com/JosefPihrt/Roslynator/pull/1031)).
 - Do not report `System.Windows.DependencyPropertyChangedEventArgs` as unused parameter ([RCS1163](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1163.md)) ([#1068](https://github.com/JosefPihrt/Roslynator/pull/1068)).
 - Fix ([RCS1032](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1032.md)) ([#1064](https://github.com/JosefPihrt/Roslynator/pull/1064)).
+- Update processing of .gloabalconfig file to prioritize file-specific diagnostic severities over global diagnostic severities. [#1066](https://github.com/JosefPihrt/Roslynator/pull/1066/files)
```
JosefPihrt, line 36:
```suggestion
- Update processing of .globalconfig file to prioritize file-specific diagnostic severities over global diagnostic severities. [#1066](https://github.com/JosefPihrt/Roslynator/pull/1066/files)
```

JosefPihrt (changes requested):

PR #1069, opened 2023-04-13
Enhance SyntaxLogicalInverter to handle CoalesceExpressions and add corresponding tests

jamesHargreaves12:
Add explicit handling of CoalesceExpressions in the SyntaxLogicalInverter class. Currently, in Debug mode the Debug.Fail on L213 will be evaluated which is a less-than-ideal experience.

This PR also adds intelligent support for the case where the RHS of the coalesce operator is either true or false.
- `x??true` is inverted to `x==false` 
- `x??false` is inverted to `x!=true`


src/CSharp.Workspaces/CSharp/SyntaxLogicalInverter.cs
```diff
@@ -208,6 +208,23 @@ internal static SyntaxLogicalInverter GetInstance(Document document)
                 {
                     return DefaultInvert(expression);
                 }
+            case SyntaxKind.CoalesceExpression:
+            {
+                var binaryExpression = (BinaryExpressionSyntax)expression;
+                if (binaryExpression.Right.Kind() == SyntaxKind.FalseLiteralExpression)
```
JosefPihrt, line 214: Please use `IsKind` instead of `Kind`. It it preferred in the codebase. It has also slightly better perf.

src/CSharp.Workspaces/CSharp/SyntaxLogicalInverter.cs
```diff
@@ -208,6 +208,23 @@ internal static SyntaxLogicalInverter GetInstance(Document document)
...
+                {
+                    // !(x ?? false) === (x != true)
+                    return BinaryExpression(SyntaxKind.NotEqualsExpression, binaryExpression.Left, LiteralExpression(SyntaxKind.TrueLiteralExpression));
```
JosefPihrt, line 217: `CSharpFactory` provides some helper methods to make these calls shorter: `CSharpFactory.NotEqualsExpression` and `CSharpFactory.TrueLiteralExpression`.

src/Tests/Analyzers.Tests/RCS1208ReduceIfNestingTests.cs
```diff
@@ -50,6 +50,126 @@ void M2()
 ");
     }
     
+    
+    [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.ReduceIfNesting)]
+    public async Task Test_InvertingCoalesceToFalse()
```
JosefPihrt, line 55: 👍 for adding tests

JosefPihrt (changes requested):

JosefPihrt (approved):

JosefPihrt:
👍

PR #1070, opened 2023-04-13
Implement FixAllProvider for RCS1014

jamesHargreaves12:
This PR introduces a new FixAllProvider for RCS1014, designed to apply code fixes sequentially in reverse order by location within the document. This approach is safe because applying a given code fix for RCS1014 will not impact the text prior to it.

The FixAllProvider is constructed to minimize its coupling to any RCS1014-specific logic, making it easier to extend for other code fixes that satisfy the following property:
- Applying a code fix for Diagnostic X at location Y would not change the location, validity or code fixes of a Diagnostics of the same type as X but that occurs earlier in the file

As an aside, when first looking at this, I tried using WellKnownFixAllProviders.BatchFixAll rather than implementing something new. On occasion, this produced garbage code. For example:
```
    public static TheoryData<string[], int[]?, int[]?, int[]?, int[]?, int?[]?, int[]> M()
    {
        var l = nameof(X);
        var r = nameof(Y);
        return new()
        {
            { new[] { l, l, r, r }, null, null, new[] { 1 }, null, null, new[] { 1, 3, 0, 2 } },
            { new[] { l, l, l, r, r, r }, null, null, null, null, null, new[] { 2 } },
        };
    }
```
Gets converted to the following:
```
    public static TheoryData<string[], int[]?, int[]?, int[]?, int[]?, int?[]?, int[]> M()
    {
        var l = nameof(X);
        var r = nameof(Y);
        return new()
        {
            { new string[] { l, l, r, r }, null, null, new[] { 1 }, null, null, new[] { 1, 3, 0, 2 } },
            { new int[] { 1 }, null, null, new[] { l, l, l, r, r, r }, null, null, null, null, null, new int[] { 2 } },
        };
    }
```
This is an issue within the merge algorithm of WellKnownFixAllProviders.BatchFixAll. After reviewing related issues on the Roslyn GitHub, it became evident that this FixAllProvider is not expected to work in all scenarios, and implementing a custom solution is preferred.


JosefPihrt:
Hi,

it would be better to open a new issue first and then open a PR.

I used your code sample (+ I used `TheoryData` from xUnit). I applied code fix "Use explicitly typed array" with option "Fix all occurrences in document". The result is as expected:

```
    public static TheoryData<string[], int[]?, int[]?, int[]?, int[]?, int?[]?, int[]> M()
    {
        var l = nameof(X);
        var r = nameof(Y);
        return new()
        {
            { new string[] { l, l, r, r }, null, null, new int[] { 1 }, null, null, new int[] { 1, 3, 0, 2 } },
            { new string[] { l, l, l, r, r, r }, null, null, null, null, null, new int[] { 2 } },
        };
    }
```

jamesHargreaves12:
I am surprised that you didn't run into the same issue. But I have opened an Issue as requested that contains a more full description of how to reproduce the issue. https://github.com/JosefPihrt/Roslynator/issues/1075

src/Analyzers.CodeFixes/CSharp/CodeFixes/UseExplicitlyOrImplicitlyTypedArrayFixAllProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+    public static readonly UseExplicitlyOrImplicitlyTypedArrayFixAllProvider Instance = new();
+    private UseExplicitlyOrImplicitlyTypedArrayFixAllProvider() { }
+
+    protected override async Task<Document> FixAllAsync(FixAllContext fixAllContext, Document document, ImmutableArray<Diagnostic> diagnostics)
+    {
+        var codeFixProvider = (UseExplicitlyOrImplicitlyTypedArrayCodeFixProvider)fixAllContext.CodeFixProvider;
+        foreach (var diag in diagnostics.OrderByDescending(d => d.Location.SourceSpan.Start))
+        {
+            document = await codeFixProvider.ApplyFixToDocument(document, diag, fixAllContext.CancellationToken);
+        }
```
JosefPihrt, line 21: What if diagnostic are nested? Will it work then?

JosefPihrt (changes requested):

src/Analyzers.CodeFixes/CSharp/CodeFixes/UseExplicitlyOrImplicitlyTypedArrayFixAllProvider.cs
JosefPihrt: 👍

src/Analyzers.CodeFixes/CSharp/CodeFixes/UseExplicitlyOrImplicitlyTypedArrayCodeFixProvider.cs
```diff
@@ -39,29 +55,40 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
         Document document = context.Document;
         Diagnostic diagnostic = context.Diagnostics[0];
 
-        switch (node)
+        string title = node switch
+        {
+            ImplicitArrayCreationExpressionSyntax => "Use explicitly typed array",
+            ArrayCreationExpressionSyntax => "Use implicitly typed array",
+            _ => "",
```
JosefPihrt, line 62: This arm is unreachable, so I would prefer to throw InvalidOperationException.

src/Analyzers.CodeFixes/CSharp/CodeFixes/UseExplicitlyOrImplicitlyTypedArrayCodeFixProvider.cs
```diff
@@ -39,29 +55,40 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
-                    break;
-                }
+            return null;
         }
+
+        return node switch
+        {
+            ImplicitArrayCreationExpressionSyntax implicitArrayCreation => await ChangeArrayTypeToExplicitAsync(document, implicitArrayCreation, cancellationToken).ConfigureAwait(false),
+            ArrayCreationExpressionSyntax arrayCreation => await ChangeArrayTypeToImplicitAsync(document, arrayCreation, cancellationToken).ConfigureAwait(false),
+            _ => null
```
JosefPihrt, line 90: This arm is unreachable, so I would prefer to throw InvalidOperationException.

JosefPihrt (changes requested):

ChangeLog.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add SECURITY.md ([#1147](https://github.com/josefpihrt/roslynator/pull/1147))
+- Add Custom FixAllProvider for [RCS1014](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1014.md) ([#1070](https://github.com/JosefPihrt/Roslynator/pull/1070)).
```
JosefPihrt, line 13:
```suggestion
- Add custom FixAllProvider for [RCS1014](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1014.md) ([#1070](https://github.com/JosefPihrt/Roslynator/pull/1070)).
```

JosefPihrt (approved):

PR #1078, opened 2023-04-23
[CLI] Add more information about the found diagnostics to the XML output file

### LINKED ISSUE #1077: [CLI] Add Description and HelpLink to the XML output of the analyze command
The `DiagnosticDescriptor` class contains additional information about the found diagnostics, such as the `Description` and `Help Link`. In my opinion, it would be useful to add these fields to the XML output of `roslynator analyze` command.

PeterKaszab:
Closes #1077

### DIFF
--- a/src/CommandLine/Xml/DiagnosticXmlSerializer.cs
+++ b/src/CommandLine/Xml/DiagnosticXmlSerializer.cs
@@ -75,6 +75,8 @@ internal static class DiagnosticXmlSerializer
             new XAttribute("Id", diagnostic.Descriptor.Id),
             new XElement("Severity", diagnostic.Severity),
             new XElement("Message", diagnostic.Descriptor.Title.ToString(formatProvider)),
+            new XElement("Description", diagnostic.Descriptor.Description.ToString(formatProvider)),
+            new XElement("HelpLink", diagnostic.Descriptor.HelpLinkUri.ToString(formatProvider)),
             filePathElement,
             locationElement);
     }
--- a/src/CommandLine/Xml/DiagnosticXmlSerializer.cs
+++ b/src/CommandLine/Xml/DiagnosticXmlSerializer.cs
@@ -86,13 +86,30 @@ internal static class DiagnosticXmlSerializer
             diagnostics
                 .GroupBy(f => f.Descriptor, DiagnosticDescriptorComparer.Id)
                 .OrderBy(f => f.Key, DiagnosticDescriptorComparer.Id)
-                .Select(f => new XElement(
-                    "Diagnostic",
-                    new XAttribute("Id", f.Key.Id),
-                    new XAttribute("Title", f.Key.Title.ToString(formatProvider)),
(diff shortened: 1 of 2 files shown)


src/CommandLine/Xml/DiagnosticXmlSerializer.cs
```diff
@@ -75,6 +75,8 @@ private static XElement SerializeDiagnostic(DiagnosticInfo diagnostic, IFormatPr
             new XAttribute("Id", diagnostic.Descriptor.Id),
             new XElement("Severity", diagnostic.Severity),
             new XElement("Message", diagnostic.Descriptor.Title.ToString(formatProvider)),
+            new XElement("Description", diagnostic.Descriptor.Description.ToString(formatProvider)),
+            new XElement("HelpLink", diagnostic.Descriptor.HelpLinkUri.ToString(formatProvider)),
```
JosefPihrt, line 79: As these metadata are same for each diagnostic descriptor I would not put it here but I would put it in the summary (method `CreateSummary`).

JosefPihrt (changes requested):

src/CommandLine/Xml/DiagnosticXmlSerializer.cs
PeterKaszab: Done.

src/CommandLine/Xml/DiagnosticXmlSerializer.cs
```diff
@@ -90,6 +90,8 @@ private static XElement CreateSummary(IEnumerable<DiagnosticInfo> diagnostics, I
                     "Diagnostic",
                     new XAttribute("Id", f.Key.Id),
                     new XAttribute("Title", f.Key.Title.ToString(formatProvider)),
+                    new XAttribute("Description", f.Key.Description.ToString(formatProvider)),
+                    new XAttribute("HelpLink", f.Key.HelpLinkUri),
```
JosefPihrt, line 94: 1) I would change it from attribute to element because a) too many attributes make it less readable b) Description can be possibly long which make it unsuitable as an attribute value.

2) Both properties should be added to output only if they are not null.

JosefPihrt (changes requested):

src/CommandLine/Xml/DiagnosticXmlSerializer.cs
PeterKaszab: What about the existing fields (Title, Count). Should I refactor them to elements as well, or should I keep them as attributes for backward compatibility?

JosefPihrt: I would keep it as they are because of backward compatibility.

PeterKaszab: Okay, I made them elements. I also moved this functionality to a separate method.

PeterKaszab:
@JosefPihrt Thank you for the quick code review!

PR #1085, opened 2023-05-02
Fix RCS1084

jamesHargreaves12:
If the types of a and b are polymorphic and we wish to null coalesce them (`a??b`), then we must cast the LHS of the null coalesce operator to the base type, or the compiler will throw an error (i.e. (Base?)a??b https://stackoverflow.com/questions/25240564/polymorphic-null-coalescing-operator

A complete example is given in the test code below.


src/Tests/Analyzers.Tests/RCS1084UseCoalesceExpressionInsteadOfConditionalExpressionTests.cs
```diff
@@ -31,6 +31,48 @@ void M()
 }
 ", source, expected);
     }
+    
+    [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.UseCoalesceExpressionInsteadOfConditionalExpression)]
+    public async Task Test_PolymorphicType()
```
JosefPihrt, line 36: Please add similar test for case when nullable context is not enabled.

src/Analyzers.CodeFixes/CSharp/Refactorings/SimplifyNullCheckRefactoring.cs
```diff
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Josef Pihrt and Contributors. Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Linq;
```
JosefPihrt, line 3: Unnecessary when proposed changes are applied.

src/Analyzers.CodeFixes/CSharp/Refactorings/SimplifyNullCheckRefactoring.cs
```diff
@@ -44,6 +45,16 @@ internal static class SimplifyNullCheckRefactoring
...
+            
+            // If the types are polymorphic then the LHS of the null coalesce must be cast to the base type.
+            var newNodeType = semanticModel.GetTypeInfo(newNode).Type;
+            var whenNullType = semanticModel.GetTypeInfo(whenNull).Type;
+            var overallType = semanticModel.GetTypeInfo(conditionalExpression).ConvertedType;
+            if (overallType != null && newNodeType != null && !newNodeType.Equals(whenNullType))
+            {
+                var typeCastStr = overallType.ToMinimalDisplayString(semanticModel, overallType.Locations.First().SourceSpan.Start);
+                newNode = CastExpression(NullableType(ParseTypeName(typeCastStr)), newNode);
+            }
```
JosefPihrt, line 57: In general, this code does what is should do but there are many small improvements.

I wrote entire block of code instead of listing each suggestion separately:

```
ITypeSymbol newNodeType = semanticModel.GetTypeSymbol(newNode);
ITypeSymbol whenNullType = semanticModel.GetTypeSymbol(whenNull);
ITypeSymbol overallType = semanticModel.GetTypeInfo(conditionalExpression).ConvertedType;

if (overallType?.SupportsExplicitDeclaration() == true
    && !SymbolEqualityComparer.Default.Equals(newNodeType, whenNullType))
{
    TypeSyntax castType = overallType.ToTypeSyntax().WithSimplifierAnnotation();

    if ((semanticModel.GetNullableContext(conditionalExpression.SpanStart) & NullableContext.AnnotationsEnabled) != 0)
        castType = NullableType(castType);

    newNode = CastExpression(
        castType,
        newNode.WithoutTrivia())
        .WithTriviaFrom(newNode);
}
```

JosefPihrt (changes requested):

JosefPihrt (approved):

PR #1091, opened 2023-05-24
Fix RCS1047 for various IAsyncEnumerable shapes

### LINKED ISSUE #1084: RCS1047 "Remove Async suffix" is incorrectly reported for IAsyncEnumerable
Original title: RCS1047 "Remove Async suffix" is incorrectly reported for IAsyncEnumerables

**Product and Version Used**: 
Roslynator 4.3.0 on Visual Studio 2022 17.5.5 64bit

**Steps to Reproduce**:
```csharp
public static class Test
{
    public static AsyncEnumerable EnumerateAsync() // <- reported here
    {
        return new AsyncEnumerable();
    }
}

public readonly struct AsyncEnumerable
{
    public IAsyncEnumerator<int> GetAsyncEnumerator(CancellationToken cancellationToken = default)
    {
        throw new NotImplementedException();
    }
}
```

**Actual Behavior**:
The `EnumerateAsync`-method's `Async`-suffix is faded, and roslynator suggests to remove the suffix. If the returned type implements `IAsyncEnumerable<>`, the analyzer isn't triggered. This might be intended or an oversight, and luckily it's not a big deal if it won't be fixed.

**Expected Behavior**:
The analyzer could check if the type implements `GetAsyncEnumerator` directly. This is another can of worms too, though, since the enumerator isn't required to implement the interface either, as long as it has  `Current` and `MoveNextAsync()`.

---

My reasoning for foregoing the interface is to avoid the enumerable/enumerator accidentally being used with 
 LINQ or boxed to be used anywhere except the callsite. As for the suffix, there are synchronous overloads in the same class, and I wanted to separate the ones returning an async enumerable.

ovska:
closes #1084

Unsure if I should add the line to changelog, let me know if it needs to be removed.

Decided to include support for duck-typed `GetAsyncEnumerator()` after all, but left the analysis at that. The fact that the enumerator can be duck typed too (and via generics as well!). The current implementation should handle 99,9%+ of cases. If the inheritance walking and method existence checks are too heavy, let me know and I'll remove them and leave just the generics/interface impl support.

To demonstrate the madness that full duck typing support could possibly have, this compiles fine:
```csharp
await foreach (var a in new C().GetTObjAsync<IAsyncEnumerator<object?>>())
{
}

class C
{
    public Obj<T> GetTObjAsync<T>()
    {
        throw new NotImplementedException();
    }
}

abstract class Obj<T>
{
    public abstract T GetAsyncEnumerator();
}
```

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 - Fix [RCS1016](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1016.md) ([#1090](https://github.com/josefpihrt/roslynator/pull/1090)).
+- Recognize more shapes of IAsyncEnumerable as being Async ([RCS1047](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1047.md)) ([#1084](https://github.com/josefpihrt/roslynator/pull/1084)).
 
 ### Fixed
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,14 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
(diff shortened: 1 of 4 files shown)


JosefPihrt (approved):

JosefPihrt:
Thanks for the PR! 👍

PR #1092, opened 2023-05-30
Fix RCS1169

jamesHargreaves12:
Currently, RCS1169 will suggest setting variables to read-only in cases which would cause compile-time errors such as the following:
```
class C
{
    int _a;
    void M2() {M(ref _a!);} 
}
```
This is because when working out what symbols have values assigned to them outside a constructor it currently does not consider symbols that have the SuppressNullableWarning suffix (`!`).

There is another minor issue is that anything inside the ref expression is not being considered by the MakeMemberReadonlyWalker.


JosefPihrt:
Nice! 👍

PR #1094, opened 2023-06-05
Fix RCS1216

jamesHargreaves12:
Currently, if an unsafe local block is located inside a for statement then one of the asserts will fail https://github.com/JosefPihrt/Roslynator/blob/1d035ba7b32fafd9057e3c02f3a903a1493ae212/src/Analyzers/CSharp/Analysis/UnnecessaryUnsafeContextAnalyzer.cs#L248

For example
```
    class C
    {
        unsafe void M()
        {
            for(int y = 0; y < 10; y ++){
                unsafe void M2() {
                    var x = 1;
                }
            }
        }
    }
```
To fix this and make the code easier to grok I refactored the 3 distinct methods of walking up the syntax tree checking for unsafe modifiers into a single loop. I also Added a new test suite for the analyzer/code fix of RCS1216 as one currently does not exist.


src/Tests/Analyzers.Tests/RCS1216UnnecessaryUnsafeContextTests.cs
```diff
@@ -0,0 +1,249 @@
...
+    class C
+    {
+        unsafe <member_signature>
+        {
+            [|unsafe|] {
+                var x = 1;
+            }
+        }
+    }
+".Replace("<member_signature>", memberSignature);
```
JosefPihrt, line 63: This "replace" pattern is not used anywhere in the unit tests. Please create separate method for each case.

src/Tests/Analyzers.Tests/RCS1216UnnecessaryUnsafeContextTests.cs
```diff
@@ -0,0 +1,249 @@
...
+        {
+            unsafe {
+                var x = 1;
+            }
+        }
+    }
+", options: Options.WithAllowUnsafe(true));
+    }
+
+
```
JosefPihrt, line 248: Unnecessary empty lines.

JosefPihrt (changes requested):
If you want to simplify the code even more, the method

```
private static void AnalyzeMemberDeclaration(SyntaxNodeAnalysisContext context)
```
can include all type declarations and member declarations. Virtually all cases except UnsafeStatement and LocalFunctionStatement.

JosefPihrt (changes requested):

src/Analyzers/CSharp/Analysis/UnnecessaryUnsafeContextAnalyzer.cs
```diff
@@ -176,104 +157,38 @@ private static void AnalyzeIndexerDeclaration(SyntaxNodeAnalysisContext context)
...
+        while (node is not null)
         {
-            SyntaxKind kind = parent.Kind();
-
-            if (kind == SyntaxKind.UnsafeStatement)
-            {
-                return true;
-            }
-            else if (kind == SyntaxKind.LocalFunctionStatement)
+            switch (node)
```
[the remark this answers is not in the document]
jamesHargreaves12: Ah, good point - I misunderstood what you were asking

JosefPihrt (approved):

PR #1095, opened 2023-06-06
Fix RCS1211

jamesHargreaves12:
When removing unnecessary else blocks we need to also check if it conflicts with any other switch blocks (If the else is located at the top level of a switch selection).
e.g. in the following example, we cannot apply RCS1211 to remove the else.
```
        switch(i)
        {
            case 0:
                if(flag)
                {
                    var z = 1;
                    return z;
                }   
                break;
            case 1:
                if(flag)
                {
                    var y = 1;
                    return y;
                }
                else
                {
                    var z = 1;
                    return z;
                }
        }
```

The logic here is identical to removing the braces around the switch statements and so I just refactored that method out into its own helper class.


src/Analyzers/CSharp/Analysis/RemoveUnnecessaryElseAnalyzer.cs
```diff
@@ -61,8 +61,15 @@ private static bool IsFixable(ElseClauseSyntax elseClause, SemanticModel semanti
...
-            return false;
+        if (elseClause.Statement is BlockSyntax elseBlock)
+        {
+            if (LocalDeclaredVariablesOverlap(elseBlock, ifBlock, semanticModel))
+                return false;
+
+            if (ifStatement.Parent is SwitchSectionSyntax { Parent: SwitchStatementSyntax switchStatement }
+                && SwitchLocallyDeclaredVariablesHelper.BlockDeclaredVariablesOverlapWithOtherSwitchSections(elseBlock, switchStatement, semanticModel))
+                return false;
+        }
```
JosefPihrt, line 72:
```suggestion
                && SwitchLocallyDeclaredVariablesHelper.BlockDeclaredVariablesOverlapWithOtherSwitchSections(elseBlock, switchStatement, semanticModel))
            {
                return false;
            }
        }
```

src/Analyzers/CSharp/Analysis/SwitchLocallyDeclaredVariablesHelper.cs
```diff
@@ -0,0 +1,47 @@
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Roslynator.CSharp.Analysis;
+
+public static class SwitchLocallyDeclaredVariablesHelper
```
JosefPihrt, line 8:
```suggestion
internal static class SwitchLocallyDeclaredVariablesHelper
```

src/Analyzers/CSharp/Analysis/SwitchLocallyDeclaredVariablesHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+            foreach (StatementSyntax statement in otherSection.Statements)
+            {
+                foreach (ISymbol symbol in semanticModel.AnalyzeDataFlow(statement)!.VariablesDeclared)
+                {
+                    if (sectionDeclaredVariablesNames.Contains(symbol.Name))
+                        return true;
+                }
+            }
+        }
+        return false;
```
JosefPihrt, line 45:
```suggestion
        }

        return false;
```

src/Analyzers/CSharp/Analysis/SwitchLocallyDeclaredVariablesHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+    }
+}
```
JosefPihrt, line 47:
```suggestion
}

```

JosefPihrt (changes requested):

JosefPihrt (approved):

PR #1098, opened 2023-06-14
Fix RCS1146

jamesHargreaves12:
If we have the following expression and the field F is a value type then we cannot apply RCS1146 as this will mean that a different overload of `==` is used (and one may not exist).
```
c != null && c.F == "someStr"
```

We know that the `==` operator exists and works correctly for in-built value types like int so the diagnostic can still be suggested for those cases.

### DIFF
--- a/src/Analyzers/CSharp/Analysis/UseConditionalAccessAnalyzer.cs
+++ b/src/Analyzers/CSharp/Analysis/UseConditionalAccessAnalyzer.cs
@@ -275,6 +275,10 @@ public sealed class UseConditionalAccessAnalyzer : BaseDiagnosticAnalyzer
                 case SyntaxKind.EqualsExpression:
                     {
                         var leftTypeSymbol = semanticModel.GetTypeSymbol(((BinaryExpressionSyntax)expression).Left);
+
+                        if (leftTypeSymbol.IsErrorType())
+                            return false;
+
                         if (leftTypeSymbol.IsValueType && !CSharpFacts.IsPredefinedType(leftTypeSymbol.SpecialType))
                         { 
                             // If the LHS is a ValueTypes then making the expression conditional would change the type of the expression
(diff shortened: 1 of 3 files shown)


src/Analyzers/CSharp/Analysis/UseConditionalAccessAnalyzer.cs
```diff
@@ -274,6 +274,14 @@ static bool ExistsImplicitConversionToBoolean(INamedTypeSymbol typeSymbol)
                 case SyntaxKind.GreaterThanOrEqualExpression:
                 case SyntaxKind.EqualsExpression:
                     {
+                        var leftTypeSymbol = semanticModel.GetTypeSymbol(((BinaryExpressionSyntax)expression).Left);
+                        if (leftTypeSymbol.IsValueType && leftTypeSymbol.SpecialType is SpecialType.None)
```
JosefPihrt, line 278: I would rather use `CSharpFacts.IsPredefinedType` instead of `is SpecialTypen.None`.

JosefPihrt (changes requested):

src/Analyzers/CSharp/Analysis/UseConditionalAccessAnalyzer.cs
```diff
@@ -274,6 +274,14 @@ static bool ExistsImplicitConversionToBoolean(INamedTypeSymbol typeSymbol)
                 case SyntaxKind.GreaterThanOrEqualExpression:
                 case SyntaxKind.EqualsExpression:
                     {
+                        var leftTypeSymbol = semanticModel.GetTypeSymbol(((BinaryExpressionSyntax)expression).Left);
+                        if (leftTypeSymbol.IsValueType && !CSharpFacts.IsPredefinedType(leftTypeSymbol.SpecialType))
```
JosefPihrt, line 278:
```suggestion
                        var leftTypeSymbol = semanticModel.GetTypeSymbol(((BinaryExpressionSyntax)expression).Left);

                        if (leftTypeSymbol.IsErrorType())
                            return false;

                        if (leftTypeSymbol.IsValueType && !CSharpFacts.IsPredefinedType(leftTypeSymbol.SpecialType))
```

JosefPihrt (changes requested):

src/Analyzers/CSharp/Analysis/UseConditionalAccessAnalyzer.cs
JosefPihrt: Include the case where the LHS expression contains invalid code.

JosefPihrt (approved):

JosefPihrt:
👍

PR #1110, opened 2023-07-11
Add docs on how to build *.nupkg packages

bdovaz:


JosefPihrt:
I'm going to finally add workflow in https://github.com/JosefPihrt/Roslynator/pull/1112 which makes this PR obsolete.

But thanks anyway.

bdovaz:
@JosefPihrt Actually, my PR is equally valid because I'm talking about making a build locally for someone who does a fork, the workflow I imagine involves publishing the package, etc.

JosefPihrt:
I'm not against adding documentation of any kind if it's helpful.

As of now, I use powershell scripts to build "Release" artifacts. When #1112 is merged all artifacts will be available through GH Actions which is convenient. All "Release" scripts will be removed.

If you want to build something locally (without creating a PR that will trigger the workflow), you can just build it inside Visual Studio or run `dotnet pack`. I think that obvious and it does not need to be documented.

If you still want to add some documentation I'm fine with it but please discuss what and how it should be documented.

PR #1120, opened 2023-07-20
Fix RCS1085

jamesHargreaves12:
The WhitespaceRemover class doesn't handle the removal of new lines from the leading whitespace correctly - the current behaviour is to throw an error on line 49.


src/Tests/Analyzers.Tests/RCS1085UseAutoPropertyTests.cs
```diff
@@ -919,6 +919,35 @@ class C
...
+
+    [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.UseAutoProperty)]
+    public async Task TestFix_WithLeadingWhitespace()
+    {
+        await VerifyDiagnosticAndFixAsync(@"
+class C
+{
+    public int [|P|]
+        {
+            get
```
JosefPihrt, line 933: I'm not able to reproduce the bug with this unit test. I'm able to reproduce it only after I add two empty lines before `get`.
```suggestion
        {


            get
```

src/CSharp/CSharp/SyntaxRewriters/WhitespaceRemover.cs
```diff
@@ -42,11 +42,16 @@ public override SyntaxTrivia VisitTrivia(SyntaxTrivia trivia)
         if (trivia.IsKind(SyntaxKind.EndOfLineTrivia))
         {
             // We can only safely remove EndOfLineTrivia if it is not proceeded by a SingleLineComment
-            int triviaIndex = trivia.Token.TrailingTrivia.IndexOf(trivia);
+            SyntaxTriviaList triviaList = (trivia.Token.TrailingTrivia.IndexOf(trivia) == -1)
+                ? trivia.Token.LeadingTrivia
+                : trivia.Token.TrailingTrivia;
+
+            int triviaIndex = triviaList.IndexOf(trivia);
+
```
JosefPihrt, line 50: It's not necessary to look twice every time. Only if the trivia is not found in the first list.

src/CSharp/CSharp/SyntaxRewriters/WhitespaceRemover.cs
```diff
@@ -42,11 +42,16 @@ public override SyntaxTrivia VisitTrivia(SyntaxTrivia trivia)
         if (trivia.IsKind(SyntaxKind.EndOfLineTrivia))
         {
             // We can only safely remove EndOfLineTrivia if it is not proceeded by a SingleLineComment
```
JosefPihrt, line 44: This comment does not make sense here. It should be moved before
```
if (!prevTrivia.IsKind(SyntaxKind.SingleLineCommentTrivia))
```

src/CSharp/CSharp/SyntaxRewriters/WhitespaceRemover.cs
```diff
@@ -42,11 +42,16 @@ public override SyntaxTrivia VisitTrivia(SyntaxTrivia trivia)
...
+                ? trivia.Token.LeadingTrivia
+                : trivia.Token.TrailingTrivia;
+
+            int triviaIndex = triviaList.IndexOf(trivia);
+
             if (triviaIndex == 0)
                 return Replacement;
 
-            SyntaxTrivia prevTrivia = trivia.Token.TrailingTrivia[triviaIndex - 1];
+            SyntaxTrivia prevTrivia = triviaList[triviaIndex - 1];
```
JosefPihrt, line 54: prevTrivia can be inlined

JosefPihrt (changes requested):

JosefPihrt (approved):

JosefPihrt:
👍

PR #1129, opened 2023-07-24
Fix RCS1179

jamesHargreaves12:
If the field is explicitly defined this can cause implicit casting on assignment. In these cases explicitly returning rather than assigning (as RCS1179 suggests) would change the return type of the function (leading to the code not compiling) - see test code for examples.

This PR adds a check to prevent the diagnostic from being suggested if the return type doesn't exactly match the explicit type of the variable.


src/Analyzers/CSharp/Analysis/UnnecessaryAssignmentAnalyzer.cs
```diff
@@ -70,6 +70,8 @@ private static void AnalyzeIfStatement(SyntaxNodeAnalysisContext context)
         if (!IsLocalDeclaredInScopeOrNonRefOrOutParameterOfEnclosingSymbol(symbol, statementsInfo.Parent, semanticModel, cancellationToken))
             return;
 
+        ITypeSymbol returnTypeSymbol = semanticModel.GetTypeInfo(expression, cancellationToken).Type;
```
JosefPihrt, line 73:
```suggestion
        ITypeSymbol returnTypeSymbol = semanticModel.GetTypeSymbol(expression, cancellationToken);
```

src/Analyzers/CSharp/Analysis/UnnecessaryAssignmentAnalyzer.cs
```diff
@@ -204,11 +207,12 @@ private static bool IsLocalDeclaredInScopeOrNonRefOrOutParameterOfEnclosingSymbo
...
 
-    private static bool IsSymbolAssignedInStatement(ISymbol symbol, StatementSyntax statement, SemanticModel semanticModel, CancellationToken cancellationToken)
+    private static bool IsSymbolAssignedInStatementWithCorrectType(ISymbol symbol, StatementSyntax statement, SemanticModel semanticModel, ITypeSymbol typeSymbol, CancellationToken cancellationToken)
     {
         SimpleAssignmentStatementInfo assignmentInfo = SyntaxInfo.SimpleAssignmentStatementInfo(statement);
 
         return assignmentInfo.Success
-            && SymbolEqualityComparer.Default.Equals(semanticModel.GetSymbol(assignmentInfo.Left, cancellationToken), symbol);
+            && SymbolEqualityComparer.Default.Equals(semanticModel.GetSymbol(assignmentInfo.Left, cancellationToken), symbol)
+            && typeSymbol.Equals(semanticModel.GetTypeInfo(assignmentInfo.Right, cancellationToken).Type);
```
JosefPihrt, line 216: Use SymbolEqualityComparer + ext. method GetTypeSymbol

JosefPihrt (changes requested):

JosefPihrt (approved):

PR #1152, opened 2023-08-06
Support more cases for RCS1151

jamesHargreaves12:
Add support for a couple of cases:
1) Casts where the type is identical :
```
int x = 1;
int y = (int)x;
```

2) Linq Cast from a type that implements IEnumerable<T> and chained to a method that receives an IEnumerable<T>.
```
Regex.Matches(s, ".*")
                .Cast<Match>()
                .Select(x => x.Value);
```


JosefPihrt:
@jamesHargreaves12 pipeline failed.

jamesHargreaves12:
Cannot reproduce the pipeline error locally.

src/Analyzers/CSharp/Analysis/RemoveRedundantCastAnalyzer.cs
```diff
@@ -238,12 +247,34 @@ public static void Analyze(SyntaxNodeAnalysisContext context, in SimpleMemberInv
...
+            return;
+
+        ITypeSymbol genericParameter;
+        if (memberAccessExpressionType.OriginalDefinition.IsIEnumerableOfT())
+        {
+            genericParameter = memberAccessExpressionType.TypeArguments[0];
+        }
+        else if (invocationExpression.Parent is not MemberAccessExpressionSyntax { Parent: InvocationExpressionSyntax chainedMethodInvocation }
+                 || semanticModel.GetSymbol(chainedMethodInvocation, cancellationToken) is not IMethodSymbol { ReceiverType: INamedTypeSymbol chainedMethodReceiverType }
+                 || chainedMethodReceiverType.OriginalDefinition.IsIEnumerableOfT() != true)
```
JosefPihrt, line 261:
```suggestion
        else if (invocationExpression.Parent is not MemberAccessExpressionSyntax { Parent: InvocationExpressionSyntax chainedMethodInvocation }
            || semanticModel.GetSymbol(chainedMethodInvocation, cancellationToken) is not IMethodSymbol { ReceiverType: INamedTypeSymbol chainedMethodReceiverType }
            || !chainedMethodReceiverType.OriginalDefinition.IsIEnumerableOfT())
```

JosefPihrt (changes requested):

PR #1153, opened 2023-08-06
Fix RCS1208

jamesHargreaves12:
One line fix to handle Charp <= 8 correctly (no "is not" expressions).


ChangeLog.md
```diff
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix [RCS0005](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0005) ([#1114](https://github.com/JosefPihrt/Roslynator/pull/1114)).
 - Fix [RCS1176](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1176.md) ([#1122](https://github.com/JosefPihrt/Roslynator/pull/1122), [#1140](https://github.com/JosefPihrt/Roslynator/pull/1140)).
 - Fix [RCS1085](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1085.md) ([#1120](https://github.com/josefpihrt/roslynator/pull/1120)).
-- Fix [RCS1208](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1208.md) ([#1119](https://github.com/JosefPihrt/Roslynator/pull/1119)).
+- Fix [RCS1208](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1208.md) ([#1119](https://github.com/JosefPihrt/Roslynator/pull/1119),[#1153](https://github.com/JosefPihrt/Roslynator/pull/1153)).
```
JosefPihrt, line 43: Please move it to "Unreleased"

JosefPihrt (changes requested):

src/CSharp.Workspaces/CSharp/SyntaxLogicalInverter.cs
```diff
@@ -128,13 +128,17 @@ internal static SyntaxLogicalInverter GetInstance(Document document)
...
                 {
                     var isExpression = (BinaryExpressionSyntax)expression;
-
-                    return IsPatternExpression(
-                        isExpression.Left,
-                        isExpression.OperatorToken.WithTrailingTrivia(Space),
-                        UnaryPattern(
-                            Token(SyntaxKind.NotKeyword).WithTrailingTrivia(isExpression.OperatorToken.TrailingTrivia),
-                            TypePattern((TypeSyntax)isExpression.Right)));
+                    string fullyQualifiedName = semanticModel.GetSymbolInfo(isExpression.Right, cancellationToken).Symbol!
```
jamesHargreaves12, line 131: Interestingly from CSharp8 onwards, there is another case of weird behaviour that means that you cannot just replace "is" with "is not". If X is the name of the type and X is also a field on the class then `is X` points to the type and `is not X` points to the field (and hence fails to compile). See the test `Test_IsWhenIsNotIsInvalid` below for an example of this situation.

More discussion of this weirdness of the CSharp language here:
- https://github.com/dotnet/roslyn/issues/50671
- https://github.com/dotnet/csharplang/discussions/4743 

I don't know of a way to detect it without manually calculating all the in-scope meanings of the symbol in question (which I believe is non-trivial). So the best workaround that I could come up with was to specify the full qualified name to avoid any ambiguity. 

Initially, I added the `.WithSimplifyAnnotation()` so that the Simplification service would trip the fully qualified name in situations where it was necessary but unfortunately, there is a bug that means that this also won't work. I opened an issue:
https://github.com/dotnet/roslyn/issues/69426 

Aside: Is there a better way of getting a fully qualified TypeSyntax for a Symbol?

JosefPihrt: I'd use `ToTypeSyntax().WithSimplifierAnnotation()`.

Btw, are you sure you didn't forget to add Simplifier annotation? Because I tried it locally with Simplifier annotation added and it works correctly (type name is simplified as expected).

jamesHargreaves12: It must not simplify the type in the case of the test added below as that does not compile (i.e. RCS1208ReduceIfNestingTests.cs line 88).

src/CSharp.Workspaces/CSharp/SyntaxLogicalInverter.cs
```diff
@@ -128,13 +128,16 @@ internal static SyntaxLogicalInverter GetInstance(Document document)
...
                 {
                     var isExpression = (BinaryExpressionSyntax)expression;
-
-                    return IsPatternExpression(
-                        isExpression.Left,
-                        isExpression.OperatorToken.WithTrailingTrivia(Space),
-                        UnaryPattern(
-                            Token(SyntaxKind.NotKeyword).WithTrailingTrivia(isExpression.OperatorToken.TrailingTrivia),
-                            TypePattern((TypeSyntax)isExpression.Right)));
+                    var rightTypeSymbol = (ITypeSymbol)semanticModel.GetSymbolInfo(isExpression.Right, cancellationToken).Symbol!;
```
JosefPihrt, line 131: Call `GetSymbol`

src/CSharp.Workspaces/CSharp/SyntaxLogicalInverter.cs
```diff
@@ -128,13 +128,16 @@ internal static SyntaxLogicalInverter GetInstance(Document document)
...
+                    return (Options.UseNotPattern)
+                        ? IsPatternExpression(
+                            isExpression.Left,
+                            isExpression.OperatorToken.WithTrailingTrivia(Space),
+                            UnaryPattern(
+                                Token(SyntaxKind.NotKeyword)
+                                    .WithTrailingTrivia(isExpression.OperatorToken.TrailingTrivia),
+                                TypePattern(rightTypeSymbol.ToTypeSyntax(SymbolDisplayFormat.FullyQualifiedFormat))))
```
JosefPihrt, line 139: It's not necessary to specify format, just call `ToTypeSyntax()`

JosefPihrt, line 139: I don't see Simplifier annotation being added here.

JosefPihrt (changes requested):

src/CSharp.Workspaces/CSharp/SyntaxLogicalInverter.cs
jamesHargreaves12: That's because it cannot be added. https://github.com/dotnet/roslyn/issues/69426

jamesHargreaves12: I beleive it is when you need the FullyQualifiedFormat as this is not the default?

JosefPihrt: The default format used in `ToTypeSyntax` uses fully qualified name (including `global` namespace).

https://github.com/JosefPihrt/Roslynator/blob/25288a7ffe53797511ee5a18203f606b046890f6/src/CSharp/CSharp/Extensions/SymbolExtensions.cs#L275

JosefPihrt: I thought about this section one more time and if RHS expression is TypeSyntax then we can just use it as it is and we don't need to obtain symbol and then get type syntax from the symbol:

```cs
    if (Options.UseNotPattern
        && isExpression.Right is TypeSyntax type)
    {
        return IsPatternExpression(
            isExpression.Left,
            isExpression.OperatorToken.WithTrailingTrivia(Space),
            UnaryPattern(
                Token(SyntaxKind.NotKeyword)
                    .WithTrailingTrivia(isExpression.OperatorToken.TrailingTrivia),
                TypePattern(type)));
    }
    else
    {
        return DefaultInvert(expression);
    }
```

JosefPihrt: Please see my newer comment regarding the entire section.

JosefPihrt (changes requested):

src/CSharp.Workspaces/CSharp/SyntaxLogicalInverter.cs
JosefPihrt: Please see my newer comment regarding this entire section.

src/CSharp.Workspaces/CSharp/SyntaxLogicalInverter.cs
JosefPihrt: Previous post is incorrect, please ignore it.

Following code should do the trick to apply Simplifier annotation only when possible:

Condition:
```cs
bool canUseSimplifier = SymbolEqualityComparer.Default.Equals(
    semanticModel.GetSpeculativeSymbolInfo(isExpression.SpanStart, isExpression.Right, SpeculativeBindingOption.BindAsExpression).Symbol,
    semanticModel.GetSpeculativeSymbolInfo(isExpression.SpanStart, isExpression.Right, SpeculativeBindingOption.BindAsTypeOrNamespace).Symbol);
```

Entire section:
```cs
var isExpression = (BinaryExpressionSyntax)expression;
ITypeSymbol rightTypeSymbol = semanticModel.GetTypeSymbol(isExpression.Right, cancellationToken)!;

TypeSyntax type = rightTypeSymbol.ToTypeSyntax();

if (SymbolEqualityComparer.Default.Equals(
    semanticModel.GetSpeculativeSymbolInfo(isExpression.SpanStart, isExpression.Right, SpeculativeBindingOption.BindAsExpression).Symbol,
    semanticModel.GetSpeculativeSymbolInfo(isExpression.SpanStart, isExpression.Right, SpeculativeBindingOption.BindAsTypeOrNamespace).Symbol))
{
    type = type.WithSimplifierAnnotation();
}

return (Options.UseNotPattern)
    ? IsPatternExpression(
        isExpression.Left,
        isExpression.OperatorToken.WithTrailingTrivia(Space),
        UnaryPattern(
            Token(SyntaxKind.NotKeyword)
                .WithTrailingTrivia(isExpression.OperatorToken.TrailingTrivia),
            TypePattern(type)))
    : DefaultInvert(expression);
```

JosefPihrt:
@jamesHargreaves12 As soon as the conflicting file is resolved we can merge this PR.

src/CSharp.Workspaces/CSharp/SyntaxLogicalInverter.cs
```diff
@@ -128,13 +128,26 @@ internal static SyntaxLogicalInverter GetInstance(Document document)
...
-                        isExpression.Left,
-                        isExpression.OperatorToken.WithTrailingTrivia(Space),
-                        UnaryPattern(
-                            Token(SyntaxKind.NotKeyword).WithTrailingTrivia(isExpression.OperatorToken.TrailingTrivia),
-                            TypePattern((TypeSyntax)isExpression.Right)));
+                    TypeSyntax type = rightTypeSymbol.ToTypeSyntax();
+
+                    if (SymbolEqualityComparer.Default.Equals(
+                            semanticModel.GetSpeculativeSymbolInfo(isExpression.SpanStart, isExpression.Right, SpeculativeBindingOption.BindAsExpression).Symbol,
+                            semanticModel.GetSpeculativeSymbolInfo(isExpression.SpanStart, isExpression.Right, SpeculativeBindingOption.BindAsTypeOrNamespace).Symbol))
```
JosefPihrt, line 137:
```suggestion
                    if (SymbolEqualityComparer.Default.Equals(
                        semanticModel.GetSpeculativeSymbolInfo(isExpression.SpanStart, isExpression.Right, SpeculativeBindingOption.BindAsExpression).Symbol,
                        semanticModel.GetSpeculativeSymbolInfo(isExpression.SpanStart, isExpression.Right, SpeculativeBindingOption.BindAsTypeOrNamespace).Symbol))
```

JosefPihrt (approved):

PR #1154, opened 2023-08-07
Fix RCS1056

jamesHargreaves12:
RCS1056 Wasn't able to handle AliasQualifiedNames before (see test for an example that currently fails).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Append `?` to nullable reference types.
 - Fix [RCS1179](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1179.md) ([#1129](https://github.com/JosefPihrt/Roslynator/pull/1129)).
 - Fix [RCS0060](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS0060.md) ([#1139](https://github.com/JosefPihrt/Roslynator/pull/1139)).
+- Fix [RCS1056](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1056.md) ([#1154](https://github.com/JosefPihrt/Roslynator/pull/1154)).
 
 ## [4.3.0] - 2023-04-24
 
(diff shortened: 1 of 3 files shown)


src/Workspaces.Common/CSharp/Refactorings/InlineAliasExpressionRefactoring.cs
```diff
@@ -68,6 +68,20 @@ public Rewriter(IAliasSymbol aliasSymbol, TypeSyntax replacement, SemanticModel
 
         public CancellationToken CancellationToken { get; }
 
+        public override SyntaxNode VisitAliasQualifiedName(AliasQualifiedNameSyntax node)
+        {
+            IAliasSymbol aliasSymbol = SemanticModel.GetAliasInfo(node.Alias, CancellationToken);
+            if (SymbolEqualityComparer.Default.Equals(aliasSymbol, AliasSymbol)
+                && Replacement is NameSyntax replacementName)
+            {
+                return SyntaxFactory.QualifiedName(replacementName, node.Name)
```
JosefPihrt, line 77: I would say that Replacement is always NameSyntax so I'd use direct cast.

JosefPihrt (changes requested):

ChangeLog.md
```diff
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Append `?` to nullable reference types.
 - Fix [RCS1179](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1179.md) ([#1129](https://github.com/JosefPihrt/Roslynator/pull/1129)).
 - Fix [RCS0060](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS0060.md) ([#1139](https://github.com/JosefPihrt/Roslynator/pull/1139)).
+- Fix [RCS1056](https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1056.md) ([#1154](https://github.com/JosefPihrt/Roslynator/pull/1154)).
```
JosefPihrt, line 48: Move to "Unreleased"

JosefPihrt (approved):

PR #1160, opened 2023-08-15
Improve RCS1097

jamesHargreaves12:
Adds support for the following case:
```
"something" + x.ToString()
```


src/Analyzers/CSharp/Analysis/RemoveRedundantToStringCallAnalysis.cs
```diff
@@ -64,10 +64,30 @@ public static void Analyze(SyntaxNodeAnalysisContext context, in SimpleMemberInv
...
+                            if (IsNotHidden(methodSymbol, containingType))
+                                return true;
+
+                            break;
+                        }
+                    case BinaryExpressionSyntax { RawKind: (int)SyntaxKind.AddExpression } addExpression:
+                        {
+                            if (addExpression.Right == expression
+                                && semanticModel.GetTypeInfo(addExpression.Left, cancellationToken).Type?.SpecialType == SpecialType.System_String)
+                                return true;
```
JosefPihrt, line 81: Add braces

src/Analyzers/CSharp/Analysis/RemoveRedundantToStringCallAnalysis.cs
```diff
@@ -64,10 +64,30 @@ public static void Analyze(SyntaxNodeAnalysisContext context, in SimpleMemberInv
...
+
+                            if (addExpression.Left == expression
+                                && semanticModel.GetTypeInfo(addExpression.Right, cancellationToken).Type?.SpecialType == SpecialType.System_String
+                                && (addExpression.Right.WalkDownParentheses() is not InvocationExpressionSyntax invocationExpression2
+                                    || semanticModel.GetMethodSymbol(invocationExpression2, cancellationToken).Name != "ToString"))
+                                return true;
```
JosefPihrt, line 87: Add braces

src/Analyzers/CSharp/Analysis/RemoveRedundantToStringCallAnalysis.cs
```diff
@@ -64,10 +64,30 @@ public static void Analyze(SyntaxNodeAnalysisContext context, in SimpleMemberInv
...
+                    case BinaryExpressionSyntax { RawKind: (int)SyntaxKind.AddExpression } addExpression:
+                        {
+                            if (addExpression.Right == expression
+                                && semanticModel.GetTypeInfo(addExpression.Left, cancellationToken).Type?.SpecialType == SpecialType.System_String)
+                                return true;
+
+                            if (addExpression.Left == expression
+                                && semanticModel.GetTypeInfo(addExpression.Right, cancellationToken).Type?.SpecialType == SpecialType.System_String
+                                && (addExpression.Right.WalkDownParentheses() is not InvocationExpressionSyntax invocationExpression2
+                                    || semanticModel.GetMethodSymbol(invocationExpression2, cancellationToken).Name != "ToString"))
```
JosefPihrt, line 86: I'm not sure how the condition `semanticModel.GetMethodSymbol(invocationExpression2, cancellationToken).Name != "ToString")` should work. I tried that is not hit in the newly added tests. Could you add a test for it?

JosefPihrt (changes requested):

src/Analyzers/CSharp/Analysis/RemoveRedundantToStringCallAnalysis.cs
jamesHargreaves12: This is included in the test that I have added
```
string s3 = o.ToString() + o[|.ToString()|];
```

We cannot remove ToString() on both sides of the addition or we change the semantic meaning. So for this case only suggest it for the one on the right.

JosefPihrt: Ok, that makes sense 👍.

src/Analyzers/CSharp/Analysis/RemoveRedundantToStringCallAnalysis.cs
```diff
@@ -64,10 +64,23 @@ public static void Analyze(SyntaxNodeAnalysisContext context, in SimpleMemberInv
...
+                        }
+                    case SyntaxKind.AddExpression:
+                        {
+                            var addExpression = (BinaryExpressionSyntax)expression.Parent;
+                            if (addExpression.Right == expression)
+                                return semanticModel.GetTypeInfo(addExpression.Left, cancellationToken).Type?.SpecialType == SpecialType.System_String;
+
+                            return semanticModel.GetTypeInfo(addExpression.Right, cancellationToken).Type?.SpecialType == SpecialType.System_String
+                                && (addExpression.Right.WalkDownParentheses() is not InvocationExpressionSyntax invocationExpression2
+                                    || semanticModel.GetMethodSymbol(invocationExpression2, cancellationToken).Name != "ToString");
```
JosefPihrt, line 82: One more thing: It would be more precise to check if the method symbol is exactly the `ToString` we are looking for, i.e. use the same check as in the `if` condition several lines above:

```
methodSymbol?.DeclaredAccessibility == Accessibility.Public
    && !methodSymbol.IsStatic
    && !methodSymbol.IsGenericMethod
    && string.Equals(methodSymbol.Name, "ToString", StringComparison.Ordinal)
    && methodSymbol.ReturnType.SpecialType == SpecialType.System_String
    && !methodSymbol.Parameters.Any()
```

JosefPihrt (changes requested):

JosefPihrt (approved):

JosefPihrt (approved):

PR #1166, opened 2023-08-17
Fix RCS1197 

jamesHargreaves12:
C# 11 adds support for multiline interpolated strings. The code fix for RCS1197 currently produced code that cannot compile.

### DIFF
--- a/src/Workspaces.Common/CSharp/Refactorings/ConvertInterpolatedStringToStringBuilderMethodRefactoring.cs
+++ b/src/Workspaces.Common/CSharp/Refactorings/ConvertInterpolatedStringToStringBuilderMethodRefactoring.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Immutable;
 using System.Text;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Roslynator.Text;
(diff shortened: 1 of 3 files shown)


src/Workspaces.Common/CSharp/Refactorings/ConvertInterpolatedStringToStringBuilderMethodRefactoring.cs
```diff
@@ -70,10 +71,25 @@ public static (SyntaxKind contentKind, string methodName, ImmutableArray<Argumen
...
                     text = StringUtility.ReplaceDoubleBracesWithSingleBrace(text);
-
-                    text = (isVerbatim)
-                        ? "@\"" + text + "\""
-                        : "\"" + text + "\"";
+                    if (content.Parent is InterpolatedStringExpressionSyntax interpolatedStringExpression
+                        && interpolatedStringExpression.StringStartToken.IsKind(SyntaxKind.InterpolatedSingleLineRawStringStartToken)
+                        )
+                    {
+                        text = "\"\"\"" + text + "\"\"\"";
```
JosefPihrt, line 78: I think it should be the same number of quotes as in the original token.

src/Workspaces.Common/CSharp/Refactorings/ConvertInterpolatedStringToStringBuilderMethodRefactoring.cs
```diff
@@ -70,10 +71,25 @@ public static (SyntaxKind contentKind, string methodName, ImmutableArray<Argumen
...
+                    }
+                    else if (content.Parent is InterpolatedStringExpressionSyntax interpolatedStringExpression2
+                        && interpolatedStringExpression2.StringStartToken.IsKind(SyntaxKind.InterpolatedMultiLineRawStringStartToken))
+                    {
+                        text = "\"\"\"\n" + text + "\n\"\"\"\n";
```
JosefPihrt, line 83: I think it should be the same number of quotes as in the original token.

JosefPihrt (changes requested):

PR #1207, opened 2023-09-22
Migrate to .NET Foundation (file banners only)

JosefPihrt:


src/Core/AccessibilityFilter.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Josef Pihrt and Contributors. Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+﻿// Copyright (c) .NET Foundation and Contributors. Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
JosefPihrt, line 1: If this format is correct I will apply it to the rest of the files.

PR #1214, opened 2023-10-01
Improve testing framework

josefpihrt:


AlmarAubel:
ohh yes this is great. I was building my own analyzer Roslynator.Testing.CSharp and tried to use a copy of AbstractCSharpDiagnosticVerifier but it was using a lot of internal functions it looks like I can now just use CompilerDiagnosticFixVerifier

josefpihrt:
@AlmarAubel @nathanpovo If you have any comments please add it to this PR. Otherwise, the PR is ready to be merged.

nathanpovo:
Thanks for your hard work @josefpihrt!

Managed to get some time to test this out. Had to add a reference to a local version of the `Testing.Common` and `Testing.CSharp.Xunit` project (it would not work with just a reference to the `Testing.CSharp.Xunit` project).

Didn't have the chance to do a proper test suite using the new package but it looks to be ok from the couple of tests I tried.

AlmarAubel:
Thank you after creating a copy of AbstractCSharpDiagnosticVerifier I could use  it on my own analyzers.

PR #1215, opened 2023-10-02
Fix globbing

josefpihrt:


src/Workspaces.Core/Extensions/Extensions.cs
```diff
@@ -18,16 +18,20 @@ namespace Roslynator;
...
         {
             SyntaxTree tree = location.SourceTree;
 
-            if (tree is not null
-                && !matcher.Match(tree.FilePath).HasMatches)
+            if (tree is not null)
             {
-                return false;
+                PatternMatchingResult result = (rootDirectoryPath is not null)
+                    ? matcher.Match(tree.FilePath, rootDirectoryPath)
```
Jozkee, line 30:
```suggestion
                    ? matcher.Match(rootDirectoryPath, tree.FilePath)
```

josefpihrt: 👍 https://github.com/dotnet/roslynator/pull/1238

PR #1233, opened 2023-10-27
Fix RCS1234

### LINKED ISSUE #1123: RCS1234 false positive
**Product and Version Used**: 4.3.0

**Steps to Reproduce**:

```csharp
public enum Foo
{
    [Obsolete]
    Bar = 1,

    Baz = 1, // RCS1234
}
```

If `Bar` is obsolete it should not be used as a replacement for 1 because that would produce compiler warning.

jakubreznak:
fixes #1123

### DIFF
--- a/src/CSharp/CSharp/CSharpUtility.cs
+++ b/src/CSharp/CSharp/CSharpUtility.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) .NET Foundation and Contributors. Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Linq;
 using System.Threading;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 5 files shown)


jakubreznak:
@dotnet-policy-service agree

src/CSharp/CSharp/CSharpUtility.cs
```diff
@@ -681,4 +681,11 @@ public static SeparatedSyntaxList<TypeParameterSyntax> GetTypeParameters(SyntaxN
                 return null;
         }
     }
+
+    public static bool IsEnumMemberObsolete(EnumMemberDeclarationSyntax enumMember)
+    {
+        return enumMember.AttributeLists
+            .SelectMany(attrList => attrList.Attributes)
+            .Any(attribute => attribute.Name.ToString() == "Obsolete");
+    }
```
josefpihrt, line 690: This implementation probably would work for most of cases but I believe it could be improved.

It's better to use a symbol and check if it has an attribute. The symbol variable is already present in the analyzer and also in the code fix provider so this should be easy.

The method to be used is `symbol.HasAttribute(MetadataName)` and static class `MetadataNames` already contains member `MetadataNames.System_ObsoleteAttribute`.

As a result the method will be more general as it will check if a symbol is marked with `Obsolete` attribute.

josefpihrt (changes requested):

src/CSharp/CSharp/CSharpUtility.cs
jakubreznak: Thanks for the suggestion. I've updated the code to use `symbol.HasAttribute(MetadataNames.System_ObsoleteAttribute)` as recommended. I really appreciate your feedback.

josefpihrt (approved):

josefpihrt:
Thanks for the contribution!

PR #1234, opened 2023-10-28
Fix refactoring 'Inline method' (RR0062)

josefpihrt:


josefpihrt:
@dotnet-policy-service agree

PR #1245, opened 2023-11-10
Add option 'roslynator_unity_code_analysis.enabled'

### LINKED ISSUE #1200: RCS1169: FP for Unity
**Product and Version Used**: 

4.5.0

**Steps to Reproduce**:

Unity has its own serialization system where if you mark for example a private field with the `[SerializeField]` attribute it allows you in the Unity IDE to edit the value of the field so it cannot be of type `readonly`.

https://docs.unity3d.com/ScriptReference/SerializeField.html

I know there are other analyzers that make the exception to ignore this rule if the field has an attribute because it may indicate that it is processed differently as in the case of Unity.

**Actual Behavior**:

Before:

```csharp
[SerializeField]
private int a;
```

After:

```csharp
[SerializeField]
private readonly int a;
```

**Expected Behavior**:
```csharp
[SerializeField]
private int a;
```

### LINKED ISSUE #1236: Unity false positive RCS1169
Related to #585, figured better make a new issue rather than necro the old one.

**Product and Version Used**: [Roslynator 2022](https://marketplace.visualstudio.com/items?itemName=josefpihrt.Roslynator2022), version 4.6.1

**Steps to Reproduce**:
- Set the following option:
  ```editorconfig
  roslynator_suppress_unity_script_methods = true
  ```
- Create a private field with the `[SerializeField]` attribute:
  ```csharp
  [SerializeField] private bool foo;
  ```

**Actual Behavior**:
- RCS1169 (Make field read-only.) gets fired. Not that surprising, as according the [docs](https://josefpihrt.github.io/docs/roslynator/configuration/) the option only affects RCS1213.
- It is necessary to explicitly suppress RCS1169.

**Expected Behavior**:
- RCS1169 doesn't get fired, as per the main developer's claim https://github.com/dotnet/roslynator/issues/585#issuecomment-1102694414.
- RCS1169 shouldn't get fired for fields with the `[SerializeField]` attribute, as these fields can be set through Unity's inspector and therefore aren't readonly. `[SerializeField]` doesn't work for readonly fields.

The name of the option doesn't really fit this case in the first place, perhaps introduce a new option along the lines of
```editorconfig
roslynator_suppress_unity_serialized_fields = true|false
```

josefpihrt:
fixes #1236


somedeveloper00:
in Unity 2022, the `roslynator_unity_code_analysis.enabled = true` successfully avoids `rcs1213` (Remove unused member declaration) but fails to avoid `rcs1169` (Make field read-only). my `.editorconfig` is as follows:
```editorconfig
[*.cs]
roslynator_unity_code_analysis.enabled = true
```

josefpihrt:
@somedeveloper00 Please open a new issue instead of leaving comment here. Thanks.

PR #1362, opened 2024-01-15
use reproducible builds

symbiogenesis:
This should help improve the integrity of the binaries.

See also: https://github.com/dotnet/reproducible-builds


symbiogenesis:
<Deterministic> is missing

symbiogenesis:
That is a fair point.

symbiogenesis:
Ok, so I removed the package reference, but kept the sdk reference

And added one missing property that they say is needed for WPF.

josefpihrt:
@symbiogenesis build failed

PR #1393, opened 2024-02-13
Fixed DeclareAsNullableCodeFixProvider for casts in variable declarations

### LINKED ISSUE #1392: Declare as nullable code fix for CS8600 breaks when casts are involved
**Product and Version Used**: 

**Steps to Reproduce**:

The following code reports CS8600 because we cast a nullable type to a non-nullable type as part of a variable declaration.
```
    private object? Get() => null;

    internal void Test() {
      var s = (string) Get(); // (1)
      string s2 = (string) Get(); // (2)
    }
```

Roslynator's code fix for CS8600 erroneously assumes the declaration type to be the problem, when it's actually the cast. So we don't get rid of the warning and in case of `var` actually introduce broken code.

**Actual Behavior**:
```csharp
      var? s = (string) Get(); // (1)
      string? s2 = (string) Get(); // (2)
```

**Expected Behavior**:
```csharp
      var s = (string?) Get(); // (1)
      string? s2 = (string?) Get(); // (2)
```

(Although in case of (2) the declaration could be adjusted by another code fix for the same diagnostic, so at least iteratively applying fixes will arrive at the same result.)

jroessel:
Casting a nullable type to a non-nullable type within a variable declaration no longer ignores the cast type and also no longer changes `var` to `var?`.

Fixes #1392.

The code got a little convoluted and the necessity to change two types at the same time required a separate method. I'm still not sure whether there's a good way of queuing two different fixes at the same time, as you can't find your syntax nodes anymore. I hope it's okay enough, though.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix analyzer [RCS0049](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0049) ([PR](https://github.com/dotnet/roslynator/pull/1386))
 - Fix analyzer [RCS1159](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1159) ([PR](https://github.com/dotnet/roslynator/pull/1390))
+- Fix code fix for [CS8600](https://josefpihrt.github.io/docs/roslynator/fixes/CS8600) changing the wrong type when casts or `var` are involved ([PR](https://github.com/dotnet/roslynator/pull/1393))
 
 ## [4.10.0] - 2024-01-24
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix analyzer [RCS0049](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0049) ([PR](https://github.com/dotnet/roslynator/pull/1386))
 - Fix analyzer [RCS1159](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1159) ([PR](https://github.com/dotnet/roslynator/pull/1390))
-- Fix code fix for [CS8600](https://josefpihrt.github.io/docs/roslynator/fixes/CS8600) changing the wrong type when casts or `var` are involved ([PR](https://github.com/dotnet/roslynator/pull/1393))
+- Fix code fix for [CS8600](https://josefpihrt.github.io/docs/roslynator/fixes/CS8600) changing the wrong type when casts or `var` are involved ([PR](https://github.com/dotnet/roslynator/pull/1393) by @jroessel)
 
 ## [4.10.0] - 2024-01-24
 
--- a/src/CodeFixes/CSharp/CodeFixes/DeclareAsNullableCodeFixProvider.cs
+++ b/src/CodeFixes/CSharp/CodeFixes/DeclareAsNullableCodeFixProvider.cs
@@ -2,6 +2,7 @@
 
 using System.Collections.Immutable;
 using System.Composition;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CodeActions;
@@ -30,7 +31,7 @@ public sealed class DeclareAsNullableCodeFixProvider : CompilerDiagnosticCodeFix
         if (!IsEnabled(diagnostic.Id, CodeFixIdentifiers.AddNullableAnnotation, context.Document, root.SyntaxTree))
             return;
 
-        if (!TryFindFirstAncestorOrSelf(root, context.Span, out SyntaxNode node, predicate: f => f.IsKind(SyntaxKind.EqualsValueClause, SyntaxKind.DeclarationExpression, SyntaxKind.SimpleAssignmentExpression)))
+        if (!TryFindFirstAncestorOrSelf(root, context.Span, out SyntaxNode node, predicate: f => f.IsKind(SyntaxKind.EqualsValueClause, SyntaxKind.DeclarationExpression, SyntaxKind.SimpleAssignmentExpression, SyntaxKind.CastExpression)))
             return;
 
         if (node is EqualsValueClauseSyntax equalsValueClause)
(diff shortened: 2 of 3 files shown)


jroessel:
@dotnet-policy-service agree company="yWorks GmbH"

src/CodeFixes/CSharp/CodeFixes/DeclareAsNullableCodeFixProvider.cs
```diff
@@ -85,4 +90,42 @@ private static void TryRegisterCodeFix(CodeFixContext context, Diagnostic diagno
...
+            {
+                NullableTypeSyntax newType = SyntaxFactory.NullableType(type.WithoutTrivia()).WithTriviaFrom(type);
+                Document changedDocument = await context.Document.ReplaceNodeAsync(type, newType, ct).ConfigureAwait(false);
+
+                // This could be in a variable declaration, so grab the new syntax root and find the newly-replaced type node
+                SyntaxNode root = await changedDocument.GetSyntaxRootAsync(ct).ConfigureAwait(false);
+                SyntaxNode insertedNewType = root.FindNode(type.Span);
+
+                // Try finding a surrounding variable declaration whose type we also have to change
+                if (insertedNewType.AncestorsAndSelf().FirstOrDefault(a => a.IsKind(SyntaxKind.EqualsValueClause)) is EqualsValueClauseSyntax equalsValueClause)
```
josefpihrt, line 111: I'm not sure about using `AncestorsAndSelf`. I would just check if parent of cast expression is assigned value (`Initializer.Value`).

src/CodeFixes/CSharp/CodeFixes/DeclareAsNullableCodeFixProvider.cs
```diff
@@ -85,4 +90,42 @@ private static void TryRegisterCodeFix(CodeFixContext context, Diagnostic diagno
...
+                    ExpressionSyntax expression = equalsValueClause.Value;
+
+                    if (equalsValueClause.IsParentKind(SyntaxKind.VariableDeclarator)
+                        && equalsValueClause.Parent.Parent is VariableDeclarationSyntax variableDeclaration
+                        && variableDeclaration.Variables.Count == 1
+                        && !variableDeclaration.Type.IsKind(SyntaxKind.NullableType)
+                        && variableDeclaration.Type is not IdentifierNameSyntax { Identifier.Text: "var" })
+                    {
+                        NullableTypeSyntax newDeclarationType = SyntaxFactory.NullableType(variableDeclaration.Type.WithoutTrivia()).WithTriviaFrom(variableDeclaration.Type);
+                        changedDocument = await changedDocument.ReplaceNodeAsync(variableDeclaration.Type, newDeclarationType, ct).ConfigureAwait(false);
```
josefpihrt, line 122: Instead of updating document twice, I would rather use something similar like:

```cs
VariableDeclarationSyntax newVariableDeclaration = variableDeclaration
    .ReplaceNode(type, newType)
    .WithType(newDeclarationType);
```

josefpihrt (changes requested):
Please update changelog.

src/CodeFixes/CSharp/CodeFixes/DeclareAsNullableCodeFixProvider.cs
jroessel: The thought was to automatically handle perhaps a ParenthesizedExpression around, but yeah, it's probably rare enough to have those in spades around initializers. Especially with the added risk of ancestor-ing through unexpected intermediary expressions. Will change.

src/CodeFixes/CSharp/CodeFixes/DeclareAsNullableCodeFixProvider.cs
jroessel: That's ... indeed much simpler. The solution with finding the correct place in the changed document _did_ feel weird and I felt there had to be a better solution. Although it only works in this case because the syntax nodes are practically neighbors, I guess. If the code fix would have to change, say, a field declaration somewhere else, there'd be no way around this dance, right?

src/CodeFixes/CSharp/CodeFixes/DeclareAsNullableCodeFixProvider.cs
```diff
@@ -85,4 +90,45 @@ private static void TryRegisterCodeFix(CodeFixContext context, Diagnostic diagno
...
+                            } variableDeclaration
+                        }
+                    }
+                    && !declarationType.IsKind(SyntaxKind.NullableType))
+                {
+                    NullableTypeSyntax newDeclarationType = SyntaxFactory.NullableType(declarationType.WithoutTrivia()).WithTriviaFrom(declarationType);
+                    VariableDeclarationSyntax newVariableDeclaration = variableDeclaration
+                        .ReplaceNode(type, newType)
+                        .WithType(newDeclarationType);
+                    return await context.Document.ReplaceNodeAsync(variableDeclaration, newVariableDeclaration, ct).ConfigureAwait(false);
```
josefpihrt, line 123:
```suggestion
                        .WithType(newDeclarationType);

                    return await context.Document.ReplaceNodeAsync(variableDeclaration, newVariableDeclaration, ct).ConfigureAwait(false);
```

src/CodeFixes/CSharp/CodeFixes/DeclareAsNullableCodeFixProvider.cs
josefpihrt: Updating multiple nodes can be tricky. And there many approaches. You can use SyntaxRewriter, method SyntaxNode.ReplaceNodes, SyntaxAnnotation to find a node in a new tree. If you are replacing members or statements you can use its index to find it in a collection etc.

ChangeLog.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix analyzer [RCS0049](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0049) ([PR](https://github.com/dotnet/roslynator/pull/1386))
 - Fix analyzer [RCS1159](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1159) ([PR](https://github.com/dotnet/roslynator/pull/1390))
+- Fix code fix for [CS8600](https://josefpihrt.github.io/docs/roslynator/fixes/CS8600) changing the wrong type when casts or `var` are involved ([PR](https://github.com/dotnet/roslynator/pull/1393))
```
josefpihrt, line 22:
```suggestion
- Fix code fix for [CS8600](https://josefpihrt.github.io/docs/roslynator/fixes/CS8600) changing the wrong type when casts or `var` are involved ([PR](https://github.com/dotnet/roslynator/pull/1393))
```
```suggestion
- Fix code fix for [CS8600](https://josefpihrt.github.io/docs/roslynator/fixes/CS8600) changing the wrong type when casts or `var` are involved ([PR](https://github.com/dotnet/roslynator/pull/1393) by @jroessel)
```

josefpihrt (approved):

josefpihrt:
Thank you for the contribution @jroessel !

PR #1413, opened 2024-02-23
Possible 'System.NullReferenceException'

mokarchi:
use conditional access


josefpihrt (dismissed):

josefpihrt:
I took a look at it again and the value must be present here, otherwise it's ok to throw. So I'm going to close this PR.

@mokarchi  Feel free to make another contribution, you can take a look at the list of issues, maybe you find something interesting here.

PR #1429, opened 2024-03-24
Fix minor typo and re-generate documentation-related files.

Treit:
Noticed a typo in the documentation for CS8403, the title for which reads:

Method with an iterator block must be 'async' to return 'IAsyncEnumerable<T<'.

Obviously that should be a `>` not a `<` character at the end.


Shidesu:
Nice catch

josefpihrt:
Thanks!

PR #1460, opened 2024-05-08
Fix analyzer RCS1246

### LINKED ISSUE #1450: RCS1246 suggestion for Last() is not usable in .NET Framework 4.8
**Product and Version Used**:  Roslynator 4.12.1 and .NET Framework 4.8

**Steps to Reproduce**:
Use the Last() method of a List object for instance.

**Actual Behavior**:
Suggestion is to use [^1] but this is not usable in .NET Framework 4.8. We did not want to suppress the issue since other suggestions are of use for us.

**Expected Behavior**:
Having either a suggestion for .NET Framework 4.8 or don't raise it for this version.

josefpihrt:
fixes #1450

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


KavenBreton-eaton:
It seems like this exact same issue is back even though we are using the latest version (4.12.4). Should I create a new issue?

PR #1462, opened 2024-05-08
Fix analyzer RCS1208

### LINKED ISSUE #1454: RCS1208 False-positive on flat if statement with multiple conditions
**Product and Version Used**:
Roslynator.Analyzers 4.12.2

**Steps to Reproduce:**
Following code or similar one containing if statement with multiple conditions, but without any nesting:
```csharp
private void Foo(string bar, int baz)
{
    if (bar == "bar" && baz == 123)  // RCS1208: Reduce 'if' nesting
    {
        var foo = "baz";
    }
}
```
Please note, that the issue doesn't occur, when there is some other code present after if statement in question:

```csharp
private void Foo(string bar, int baz)
{
    if (bar == "bar" && baz == 123)
    {
        var foo = "baz";
    }
    var thisDoes = "something preventing RCS1208 in if statement above";
}
```

**Actual Behavior**:

RCS1208 is raised, suggesting rather counterproductive change to:
```csharp
private void Foo(string bar, int baz)
{
    if (bar != "bar" || baz != 123)
    {
        return;
    }
    var foo = "baz";
}

```
**Expected Behavior**:
Do not raise RCS1208 in case of flat if statements with multiple conditions.

josefpihrt:
fixes #1454

### DIFF
--- a/src/Common/CSharp/Analysis/ReduceIfNesting/ReduceIfNestingAnalysis.cs
+++ b/src/Common/CSharp/Analysis/ReduceIfNesting/ReduceIfNestingAnalysis.cs
@@ -113,7 +113,7 @@ internal static class ReduceIfNestingAnalysis
             return Success(jumpKind, parent);
         }
 
-        if (!IsFixable(ifStatement, statements, ref jumpKind))
+        if (!IsFixable(ifStatement, statements, options, ref jumpKind))
             return Fail(node);
 
         switch (parentKind)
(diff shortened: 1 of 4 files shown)


josefpihrt:
I'm not sure that I'm following you.

BenjaminBrienen:
It suggests to invert that final if block even though it wouldn't improve readability. This PR seems to add the concept of a "last if-block" and allows them.

josefpihrt:
Tbh, I was working on this fix couple of months ago and the only thing I recall is that fix is almost impossible so I gave up 🙂

Issue #1492, opened 2024-07-16
RCS1077 for OrderBy(x => x)


bricelam:
RCS1077 should recommend replacing `.OrderBy(x => x)` with `.Order()`.

josefpihrt:
@BenjaminBrienen You are more than welcome to do so!

Please see this file where you can a reference to the analyzer and cod fix provider:

https://github.com/dotnet/roslynator/blob/6a0519c09bc59186e3e3ee912caa099dffdf043e/src/Tests/Analyzers.Tests/RCS1077OptimizeLinqMethodCallTests.cs#L11

Also, following file can be useful even though you are not adding new analyzer:

https://josefpihrt.github.io/docs/roslynator/how-to-add-new-analyzer

BenjaminBrienen:
![image](https://github.com/user-attachments/assets/9b78da78-570e-4e3f-921a-d88b1e3a2212)
I love it when I can just `git clone` a repository and all the tests pass! (except the test I just added)

BenjaminBrienen:
Analysis is working, working on code fix

BenjaminBrienen:
Apparently Order() only exists new newer versions of .NET. The code fix works though!

BenjaminBrienen:
I found something that I could use
https://github.com/dotnet/roslynator/pull/1157#discussion_r1295714425

PR #1520, opened 2024-09-14
Implement LINQ optimization: OrderBy with Identity => Order

BenjaminBrienen:
Solves #1492 

TODO:
- Unit test fails because the unit test is run with a version of .NET Core where `System.Linq` does not yet implement `IEnumerable<T>.Order()`.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


BenjaminBrienen:
@josefpihrt Would you be able to point me in the right direction? Maybe an example of an analyzer/code fix that only activates when using at least a certain version of .NET?

BenjaminBrienen:
@dotnet-policy-service agree

BenjaminBrienen:
@dotnet-policy-service agree

BenjaminBrienen:
@dotnet-policy-service agree

PR #1524, opened 2024-09-15
RCS1140 - Don't require user to document exception types if caught in same method

Qluxzz:
There was no tests for RCS1140 so I added one based on the documentation for the rule.

When I was testing this rule I was not expecting it to require me to document exception types thrown but caught in the same method.

An example from a unit test I created
```csharp
public void Foo(object parameter)
{
    try 
    {
        if (parameter == null)
            throw new ArgumentNullException(nameof(parameter));
    }
    catch (ArgumentNullException) {}
}
```
Here the exception is caught in the method but the analyzer for RCS1140 still require it to be documented on the method.

This PR checks if the exception type thrown in the method is also caught in the same method, and then removes the requirement to document it for the method.


Qluxzz:
@dotnet-policy-service agree

src/Common/CSharp/Analysis/AddExceptionToDocumentationComment/AddExceptionToDocumentationCommentAnalysis.cs
```diff
@@ -216,4 +220,37 @@ private static bool InheritsFromException(ITypeSymbol typeSymbol, INamedTypeSymb
             && typeSymbol.BaseType?.IsObject() == false
             && typeSymbol.InheritsFrom(exceptionSymbol);
     }
+
+    private static bool IsExceptionTypeCaughtByCatch(TryStatementSyntax tryStatementSyntax, IdentifierNameSyntax exceptionType)
+        => tryStatementSyntax.Catches.Any(x => x.Declaration?.Type is IdentifierNameSyntax ins1 && ins1.Identifier.Value == exceptionType.Identifier.Value);
```
josefpihrt, line 225: - Please rename x to something more meaningful.
- The comparison should be done using `SymbolEqualityComparer` (which means that you need semantic model for it).
- Consider if it wouldn't be more readable to inline this method in the `IsExceptionTypeCaughtInMethod` method.

josefpihrt (changes requested):
Please update changelog.

src/Common/CSharp/Analysis/AddExceptionToDocumentationComment/AddExceptionToDocumentationCommentAnalysis.cs
```diff
@@ -216,4 +220,37 @@ private static bool InheritsFromException(ITypeSymbol typeSymbol, INamedTypeSymb
...
+
+    /// <summary>
+    /// Walk upwards from throw statement and find all try statements in method and see if any of them catches the thrown exception type
+    /// </summary>
+    private static bool IsExceptionTypeCaughtInMethod(SyntaxNode node, ExpressionSyntax expressionSyntax)
+    {
+        if (expressionSyntax is not ObjectCreationExpressionSyntax x || x.Type is not IdentifierNameSyntax exceptionType)
```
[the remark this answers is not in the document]
Qluxzz: I now instead take the INamedTypeSymbol for the exception directly, since we'd already computed it outside of this method.

That meant that this entire section was removed.

src/Common/CSharp/Analysis/AddExceptionToDocumentationComment/AddExceptionToDocumentationCommentAnalysis.cs
Qluxzz: I inlined the IsExceptionTypeCaughtInMethod method, this in combination with the other changes made this method a whole lot simpler.

ChangeLog.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix analyzer [RCS1181](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1181) ([PR](https://github.com/dotnet/roslynator/pull/1526))
 - Fix analyzer [RCS0005](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0005) ([PR](https://github.com/dotnet/roslynator/pull/1533))
 - Fix analyzer [RCS1181](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1181) ([PR](https://github.com/dotnet/roslynator/pull/1534))
+- Fix analyzer [RCS1140](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1140) ([PR](https://github.com/dotnet/roslynator/pull/1524))
```
josefpihrt, line 23: Please move it to "Unreleased"

src/Common/CSharp/Analysis/AddExceptionToDocumentationComment/AddExceptionToDocumentationCommentAnalysis.cs
```diff
@@ -216,4 +219,32 @@ private static bool InheritsFromException(ITypeSymbol typeSymbol, INamedTypeSymb
...
+
+    /// <summary>
+    /// Walk upwards from throw statement and find all try statements in method and see if any of them catches the thrown exception type
+    /// </summary>
+    private static bool IsExceptionTypeCaughtInMethod(SyntaxNode node, ITypeSymbol exceptionSymbol, SemanticModel semanticModel, CancellationToken cancellationToken)
+    {
+        SyntaxNode parent = node.Parent;
+        while (parent is not null)
+        {
+            if (parent is TryStatementSyntax tryStatement && tryStatement.Catches.Any(catchClause => exceptionSymbol == semanticModel.GetTypeSymbol(catchClause.Declaration?.Type, cancellationToken)))
```
josefpihrt, line 231: Please use `SymbolEqualityComparer` instead of `==` operator

josefpihrt (changes requested):

josefpihrt (dismissed):
Great PR! Thanks for the contribution.

josefpihrt (approved):

PR #1535, opened 2024-09-21
Support duck-typed awaitables and task-like types for Task/Async-related analyzers

### LINKED ISSUE #1529: Task/Async-related analyzers are not aware of custom task types and duck-typed awaitables (affects RCS1046, RCS1047, RCS1090, RCS1229, and RCS1261)
### Product and Version Used
VS extension version 4.12.5

### Background
C# supports `await`ing any expression whose type has [a specific shape](https://github.com/dotnet/roslyn/blob/main/docs/features/task-types.md) ([C# spec](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12982-awaitable-expressions)).
Additionally, `async` methods can have their return type be any "task-like" type, which is [not too complex](https://github.com/dotnet/roslyn/blob/main/docs/features/task-types.md#builder-type) to implement in user code ([C# spec](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#1515-async-functions)).

Currently, as far as I've been able to tell, analyzers like RCS1047 don't check for those shapes and thus end up misfiring (or not firing) on custom task types.

### Example/reproduction code
```csharp
using System;
using System.IO;
using System.Threading.Tasks;
using System.Runtime.CompilerServices;

#pragma warning disable CA1822 // Mark members as static
#pragma warning disable CA1050 // Declare types in namespaces
#pragma warning disable CS9113 // Parameter is unread.
#nullable disable

class CustomTask
{
    // class must have a (visible) instance or extension GetAwaiter method (returning an awaiter type) to be awaitable
    public Awaiter GetAwaiter() => default;
    public CustomTask DoSomething() => this; // expected RCS1046 (if enabled)
    public CustomTask DoSomethingAsync() => this; // expected no RCS1047
    public async Task ClassicallyAsync() => await DoSomethingAsync(); // expected RCS1090 (if enabled)
    public Task ExpectedlyAsync() => ClassicallyAsync(); // no RCS1047 as expected (since Task is known)

    // must return an awaitable type
    public ConfiguredAwaitable ConfigureAwait(bool continueOnCapturedContext) => new ConfiguredAwaitable(this);
}

interface ICustomTask
{
    CustomTask DoAsync(); // expected no RCS1047
}

#region Awaiter shape
public struct Awaiter : INotifyCompletion
{
    public bool IsCompleted => false;
    public void OnCompleted(Action continuation) { }
    public void GetResult() { }
}

public struct Awaiter<T> : INotifyCompletion
{
    public bool IsCompleted => false;
    public void OnCompleted(Action continuation) { }
    public T GetResult() => default;
}
#endregion

// This attribute is required for the type to be a valid return type for an async method
[AsyncMethodBuilder(typeof(CustomTask<>))]
class CustomTask<T>
{
    //public async CustomTask Nope() { } // CS1983
    public CustomTask(T t) { }
    public Awaiter<T> GetAwaiter() => default;

    public CustomTask<int> Foo()
    {
        using FileStream fs = default!; // expected RCS1261
        return new CustomTask<int>(1);
    }

    public async CustomTask<int> FooAsync()
    {
        using FileStream fs = default!; // get RCS1261 as expected (since it checks if method is async)
        if (fs.CanRead)
        {
            return await new CustomTask<int>(1).ConfigureAwait(false);
        }
        else
        {
            return 2;
        }
    }

    public async Task Bar()
    {
        await Foo(); // expected RCS1090 (if enabled)
    }

    public CustomTask<int> Baz() // expected RCS1229
    {
        using (default(IDisposable))
        {
            return new CustomTask<int>(1);
        }
    }

    #region Async method builder shape
    public static CustomTask<T> Create() => default!;
    public CustomTask<T> Task { get; set; } = null!;
    public void SetStateMachine(IAsyncStateMachine stateMachine) { }
    public void SetException(Exception ex) { }
    public void SetResult(T result) { }
    public void Start<TStateMachine>(ref TStateMachine stateMachine)
        where TStateMachine : IAsyncStateMachine
    { }
    public void AwaitOnCompleted<TAwaiter, TStateMachine>(ref TAwaiter awaiter, ref TStateMachine stateMachine)
        where TAwaiter : INotifyCompletion
        where TStateMachine : IAsyncStateMachine
    { }
    public void AwaitUnsafeOnCompleted<TAwaiter, TStateMachine>(ref TAwaiter awaiter, ref TStateMachine stateMachine)
        where TAwaiter : ICriticalNotifyCompletion
        where TStateMachine : IAsyncStateMachine
    { }
    #endregion
}

struct ConfiguredAwaitable
{
    public ConfiguredAwaitable(CustomTask task) { }

    public Awaiter GetAwaiter() => default;
}

static class Extensions
{
    public static CustomTask<T> ConfigureAwait<T>(this CustomTask<T> task, bool continueOnCapturedContext)
    {
        return task; // just needs to be awaitable
    }
}

```

I have a [branch](https://github.com/dotnet/roslynator/compare/main...Govorunb:roslynator:main) ready with fixes, let me know if you're cool with a PR 👍

Govorunb:
### Description
These changes add the same logic for checking [duck-typed awaitables](https://github.com/dotnet/roslyn/blob/7b7951aa13c50ad768538e58ed3805898b058928/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Extensions/ISymbolExtensions.cs#L577-L639) and [task-like types](https://github.com/dotnet/roslyn/blob/c4203b867e9c0287cabb0a0674bfca096c08fd3e/src/Compilers/CSharp/Portable/Symbols/TypeSymbolExtensions.cs#L1778) that is currently in Roslyn.

This resolves #1529 - specifically, these analyzers/refactorings become aware of custom awaitables/custom task-like types:
- RCS1046/RCS1047 (async/sync method name should/should not end in `Async`)
- RCS1090 (use/remove `ConfigureAwait(false)`)
- RCS1229/RCS1174 (use/remove `async`/`await`)
- RCS1261 (dispose asynchronously)
- RR0209 (remove `async`/`await`)

### Remarks
RCS1090's behaviour is less obvious here - it seems there is no official stance for duck-typing `ConfigureAwait` (even [Roslyn](https://github.com/dotnet/roslyn/blob/149c24c648c2211ca580539d43dc97e118b80658/src/Features/Core/Portable/Completion/Providers/AbstractAwaitCompletionProvider.cs#L71) isn't [perfectly consistent](https://github.com/dotnet/roslyn/blob/149c24c648c2211ca580539d43dc97e118b80658/src/Compilers/Core/Portable/InternalUtilities/YieldAwaitableExtensions.cs#L19)), but I'm of the opinion that if the rest of the ducks are quacking, it's better to join the flock 😄 The result is that these checks are relaxed from the `Task`-specific ``ConfiguredTaskAwaitable(`1)`` return type to any awaitable return type.

I noticed there were some checks for [WinRT async interfaces](https://github.com/dotnet/roslynator/blob/df7ec0ec6ad693b10f14cc5ccd272e63b14f8016/src/Core/SymbolUtility.cs#L653), these are covered as there's an [extension `GetAwaiter()` method](https://learn.microsoft.com/en-us/dotnet/api/system.windowsruntimesystemextensions.getawaiter?view=dotnet-uwp-10.0#overloads) available on those types.

I'm not experienced with analyzers and their edge cases so it's possible I've missed writing some obvious tests, I'd love to get your feedback.

### DIFF
--- a/src/Analyzers.CodeFixes/CSharp/SyntaxRewriters/UseAsyncAwaitRewriter.cs
+++ b/src/Analyzers.CodeFixes/CSharp/SyntaxRewriters/UseAsyncAwaitRewriter.cs
@@ -24,14 +24,14 @@ internal sealed class UseAsyncAwaitRewriter : SkipFunctionRewriter
 
     public bool KeepReturnStatement { get; }
 
-    public static UseAsyncAwaitRewriter Create(IMethodSymbol methodSymbol)
+    public static UseAsyncAwaitRewriter Create(IMethodSymbol methodSymbol, SemanticModel semanticModel, int position)
     {
         ITypeSymbol returnType = methodSymbol.ReturnType.OriginalDefinition;
 
         var keepReturnStatement = false;
 
-        if (returnType.EqualsOrInheritsFrom(MetadataNames.System_Threading_Tasks_ValueTask_T)
-            || returnType.EqualsOrInheritsFrom(MetadataNames.System_Threading_Tasks_Task_T))
+        if (returnType is INamedTypeSymbol { Arity: 1 }
+            && returnType.IsAwaitable(semanticModel, position))
         {
             keepReturnStatement = true;
         }
--- a/src/Analyzers/CSharp/Analysis/UseAsyncAwaitAnalyzer.cs
+++ b/src/Analyzers/CSharp/Analysis/UseAsyncAwaitAnalyzer.cs
@@ -57,7 +57,7 @@ public sealed class UseAsyncAwaitAnalyzer : BaseDiagnosticAnalyzer
 
         IMethodSymbol methodSymbol = context.SemanticModel.GetDeclaredSymbol(methodDeclaration, context.CancellationToken);
 
-        if (!SymbolUtility.IsAwaitable(methodSymbol.ReturnType))
+        if (!methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, methodDeclaration.SpanStart))
             return;
 
         if (IsFixable(body, context))
@@ -81,7 +81,7 @@ public sealed class UseAsyncAwaitAnalyzer : BaseDiagnosticAnalyzer
 
         IMethodSymbol methodSymbol = context.SemanticModel.GetDeclaredSymbol(localFunction, context.CancellationToken);
 
-        if (!SymbolUtility.IsAwaitable(methodSymbol.ReturnType))
+        if (!methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, localFunction.SpanStart))
             return;
 
         if (IsFixable(body, context))
@@ -101,7 +101,7 @@ public sealed class UseAsyncAwaitAnalyzer : BaseDiagnosticAnalyzer
         if (context.SemanticModel.GetSymbol(simpleLambda, context.CancellationToken) is not IMethodSymbol methodSymbol)
             return;
 
-        if (!SymbolUtility.IsAwaitable(methodSymbol.ReturnType))
+        if (!methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, simpleLambda.SpanStart))
             return;
 
         if (IsFixable(body, context))
@@ -121,7 +121,7 @@ public sealed class UseAsyncAwaitAnalyzer : BaseDiagnosticAnalyzer
         if (context.SemanticModel.GetSymbol(parenthesizedLambda, context.CancellationToken) is not IMethodSymbol methodSymbol)
             return;
 
-        if (!SymbolUtility.IsAwaitable(methodSymbol.ReturnType))
+        if (!methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, parenthesizedLambda.SpanStart))
             return;
 
         if (IsFixable(body, context))
@@ -143,7 +143,7 @@ public sealed class UseAsyncAwaitAnalyzer : BaseDiagnosticAnalyzer
         if (context.SemanticModel.GetSymbol(anonymousMethod, context.CancellationToken) is not IMethodSymbol methodSymbol)
             return;
 
-        if (!SymbolUtility.IsAwaitable(methodSymbol.ReturnType))
+        if (!methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, anonymousMethod.SpanStart))
             return;
 
         if (IsFixable(body, context))
--- a/src/Analyzers/CSharp/Analysis/UseAsyncAwaitAnalyzer.cs
+++ b/src/Analyzers/CSharp/Analysis/UseAsyncAwaitAnalyzer.cs
@@ -57,7 +57,7 @@ public sealed class UseAsyncAwaitAnalyzer : BaseDiagnosticAnalyzer
 
         IMethodSymbol methodSymbol = context.SemanticModel.GetDeclaredSymbol(methodDeclaration, context.CancellationToken);
 
-        if (!methodSymbol.ReturnType.IsTaskType(context.SemanticModel, methodDeclaration.SpanStart))
+        if (!methodSymbol.ReturnType.IsTaskType() || !methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, body.SpanStart))
             return;
 
         if (IsFixable(body, context))
@@ -81,7 +81,7 @@ public sealed class UseAsyncAwaitAnalyzer : BaseDiagnosticAnalyzer
 
         IMethodSymbol methodSymbol = context.SemanticModel.GetDeclaredSymbol(localFunction, context.CancellationToken);
 
-        if (!methodSymbol.ReturnType.IsTaskType(context.SemanticModel, localFunction.SpanStart))
+        if (!methodSymbol.ReturnType.IsTaskType() || !methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, body.SpanStart))
             return;
 
         if (IsFixable(body, context))
@@ -101,7 +101,7 @@ public sealed class UseAsyncAwaitAnalyzer : BaseDiagnosticAnalyzer
         if (context.SemanticModel.GetSymbol(simpleLambda, context.CancellationToken) is not IMethodSymbol methodSymbol)
             return;
 
-        if (!methodSymbol.ReturnType.IsTaskType(context.SemanticModel, simpleLambda.SpanStart))
+        if (!methodSymbol.ReturnType.IsTaskType() || !methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, body.SpanStart))
             return;
 
         if (IsFixable(body, context))
@@ -121,7 +121,7 @@ public sealed class UseAsyncAwaitAnalyzer : BaseDiagnosticAnalyzer
         if (context.SemanticModel.GetSymbol(parenthesizedLambda, context.CancellationToken) is not IMethodSymbol methodSymbol)
             return;
 
-        if (!methodSymbol.ReturnType.IsTaskType(context.SemanticModel, parenthesizedLambda.SpanStart))
+        if (!methodSymbol.ReturnType.IsTaskType() || !methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, body.SpanStart))
             return;
 
         if (IsFixable(body, context))
@@ -143,7 +143,7 @@ public sealed class UseAsyncAwaitAnalyzer : BaseDiagnosticAnalyzer
         if (context.SemanticModel.GetSymbol(anonymousMethod, context.CancellationToken) is not IMethodSymbol methodSymbol)
             return;
 
-        if (!methodSymbol.ReturnType.IsTaskType(context.SemanticModel, anonymousMethod.SpanStart))
+        if (!methodSymbol.ReturnType.IsTaskType() || !methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, body.SpanStart))
             return;
 
         if (IsFixable(body, context))
--- a/src/Analyzers/CSharp/Analysis/UseAsyncAwaitAnalyzer.cs
+++ b/src/Analyzers/CSharp/Analysis/UseAsyncAwaitAnalyzer.cs
@@ -57,7 +57,7 @@ public sealed class UseAsyncAwaitAnalyzer : BaseDiagnosticAnalyzer
 
         IMethodSymbol methodSymbol = context.SemanticModel.GetDeclaredSymbol(methodDeclaration, context.CancellationToken);
 
-        if (!methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, methodDeclaration.SpanStart))
+        if (!methodSymbol.ReturnType.IsTaskType(context.SemanticModel, methodDeclaration.SpanStart))
             return;
 
         if (IsFixable(body, context))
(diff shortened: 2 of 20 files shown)


Govorunb:
@dotnet-policy-service agree

josefpihrt:
@Govorunb build failed

src/Analyzers/CSharp/Analysis/UseAsyncAwaitAnalyzer.cs
```diff
@@ -57,7 +57,7 @@ private static void AnalyzeMethodDeclaration(SyntaxNodeAnalysisContext context)
 
         IMethodSymbol methodSymbol = context.SemanticModel.GetDeclaredSymbol(methodDeclaration, context.CancellationToken);
 
-        if (!SymbolUtility.IsAwaitable(methodSymbol.ReturnType))
+        if (!methodSymbol.ReturnType.IsTaskType() || !methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, body.SpanStart))
```
josefpihrt, line 60: Shouldn't the operator be `&&`?

Currently if `IsTaskType` returns true the second condition is checked which does not make sense (We already know it's awaitable since it's Task).

src/Core/SymbolUtility.cs
```diff
@@ -623,53 +623,57 @@ public static ulong GetEnumValueAsUInt64(object value, INamedTypeSymbol enumType
...
-                {
-                    case "IAsyncAction":
-                    case "IAsyncActionWithProgress`1":
-                    case "IAsyncOperation`1":
-                    case "IAsyncOperationWithProgress`2":
-                        return true;
-                }
-            }
+        var awaiterTypeDefinition = getAwaiter.ReturnType.OriginalDefinition as INamedTypeSymbol;
+        if (awaiterTypeDefinition is null or { SpecialType: SpecialType.System_Void })
```
josefpihrt, line 654: Same as previous comment: Could you invert the condition and use `SpecialType.None`?

src/Core/SymbolUtility.cs
```diff
@@ -695,4 +699,33 @@ public static bool IsAwaitable(ITypeSymbol typeSymbol, bool shouldCheckWindowsRu
...
+            || definition.HasMetadataName(in MetadataNames.System_Threading_Tasks_Task_T))
+        {
+            return true;
+        }
+
+        // Roslyn checks arity, see https://github.com/dotnet/roslyn/blob/c4203b867e9c0287cabb0a0674bfca096c08fd3e/src/Compilers/CSharp/Portable/Symbols/TypeSymbolExtensions.cs#L1844
+        if (definition.Arity > 1)
+            return false;
+
+        return definition.HasAttribute(in MetadataNames.System_Runtime_CompilerServices_AsyncMethodBuilderAttribute);
```
josefpihrt, line 729: Could you incorporate first condition into return statement so it's more concise?

src/Core/SymbolUtility.cs
```diff
@@ -623,53 +623,57 @@ public static ulong GetEnumValueAsUInt64(object value, INamedTypeSymbol enumType
...
+    /// Determines if the symbol is an awaitable type.<br/>
+    /// Any type can be <see langword="await"/>ed if it has an instance or extension <c>GetAwaiter</c> method that returns a correctly-shaped awaiter type.
+    /// </summary>
+    /// <remarks>
+    /// For more information, see the <see href="https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12982-awaitable-expressions">C# language specification</see>.
+    /// </remarks>
+    /// <param name="semanticModel">Used to determine whether a candidate <c>GetAwaiter</c> method is accessible at the given <paramref name="position"/>.</param>
+    /// <param name="position">The position of the token at which the <paramref name="symbol"/> is used.</param>
+    /// <returns>A <see cref="bool"/> indicating whether the symbol is an awaitable type or a method that returns one.</returns>
+    public static bool IsAwaitable(this ISymbol? symbol, SemanticModel semanticModel, int position)
```
josefpihrt, line 636: Could you explain why do you remove the code that checks if a type is well-known awaitable type. I think we should keep it. It's better to check for it before we get into more demanding analysis if a type duck-typed awaitable type.

josefpihrt (changes requested):

josefpihrt:
@Govorunb Please update changelog.

src/Analyzers.CodeFixes/CSharp/SyntaxRewriters/UseAsyncAwaitRewriter.cs
```diff
@@ -24,14 +24,14 @@ private UseAsyncAwaitRewriter(bool keepReturnStatement)
...
 
         var keepReturnStatement = false;
 
-        if (returnType.EqualsOrInheritsFrom(MetadataNames.System_Threading_Tasks_ValueTask_T)
-            || returnType.EqualsOrInheritsFrom(MetadataNames.System_Threading_Tasks_Task_T))
+        if (returnType is INamedTypeSymbol { Arity: 1 }
+            && returnType.IsAwaitable(semanticModel, position))
         {
             keepReturnStatement = true;
         }
```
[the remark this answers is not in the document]
Govorunb: Fixed, thanks.

src/Core/SymbolUtility.cs
```diff
@@ -623,53 +623,57 @@ public static ulong GetEnumValueAsUInt64(object value, INamedTypeSymbol enumType
...
     {
-        INamedTypeSymbol? namedTypeSymbol = GetPossiblyAwaitableType(typeSymbol);
+        ITypeSymbol? typeSymbol = (symbol as ITypeSymbol) ?? (symbol as IMethodSymbol)?.ReturnType;
 
-        if (namedTypeSymbol is null)
+        if (typeSymbol is null or { SpecialType: SpecialType.System_Void })
```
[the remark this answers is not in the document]
Govorunb: `GetAwaiter` can be an extension method (as is the case with WinRT), so something like this would actually work:
```csharp
public static TaskAwaiter GetAwaiter(this int millis)
{
    return Task.Delay(millis).GetAwaiter();
}

public static async Task Example()
{
    await 2000;
}
```
Not going to vouch for that as particularly great code... but it's valid and other variations can exist.

src/Analyzers/CSharp/Analysis/UseAsyncAwaitAnalyzer.cs
Govorunb: Making a type "task-like" (i.e. applying the `AsyncMethodBuilder` attribute) doesn't actually require being awaitable, though I agree having a non-awaitable task-like type isn't very useful. Maybe there's some esoteric use for compilers like async state machines with custom behaviour, but that's way out of my depth.
```csharp
async TaskLike M()
{
    //await TaskLike; // no GetAwaiter
    await Task.CompletedTask;
}
```
For this analyzer in particular, we can't apply it unless the type is also awaitable.
```csharp
TaskLike M2() // can apply `async`, but...
{
    return TaskLike.Create(); // can't transform to `await ...`
}
```

src/Core/SymbolUtility.cs
Govorunb: Done.

src/Core/SymbolUtility.cs
Govorunb: True, thank you for catching that. Let me know if the new code is good enough for perf.

src/Core/SymbolUtility.cs
Govorunb: This case does apply - good catch!

src/Core/SymbolUtility.cs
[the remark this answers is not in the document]
josefpihrt: Interesting.

src/Analyzers/CSharp/Analysis/UseAsyncAwaitAnalyzer.cs
josefpihrt: Oh, I see what you mean here.

I propose following change to ensure that if a type is well-known awaitable type we should bail out as early as possible:

Encapsulate this condition
```
methodSymbol.ReturnType.IsTaskLikeType()
    && methodSymbol.ReturnType.IsAwaitable(context.SemanticModel, context.Node.SpanStart))
```

into a method (name is just a suggestion):

```
SymbolUtility.HasAwaitableReturnType(IMethodSymbol ...
```

And ensure that condition

```
typeSymbol.ContainingNamespace.HasMetadataName(in MetadataNames.System_Threading_Tasks)
    && (typeSymbol.EqualsOrInheritsFrom(in MetadataNames.System_Threading_Tasks_Task)
        || typeSymbol.MetadataName is "ValueTask" or "ValueTask`1")
```

will be called either in that method or in `IsAwaitable`.

src/Core/SymbolUtility.cs
```diff
@@ -695,4 +709,52 @@ public static bool IsAwaitable(ITypeSymbol typeSymbol, bool shouldCheckWindowsRu
...
+
+    /// <summary>
+    /// Determines if the type is a task type - that is, <see cref="Task"/>, <see cref="Task{TResult}"/>, or a <see href="https://github.com/dotnet/roslyn/blob/main/docs/features/task-types.md">task-like type</see>.<br/>
+    /// Only task types (and <see langword="void"/>) can be the return type of a method marked with the <see langword="async"/> keyword.
+    /// </summary>
+    /// <remarks>
+    /// For more information, see the <see href="https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#15151-general">C# language specification</see>.
+    /// </remarks>
+    /// <returns>A <see cref="bool"/> indicating whether the type is <see cref="Task"/>, <see cref="Task{TResult}"/>, or a <see href="https://github.com/dotnet/roslyn/blob/main/docs/features/task-types.md">task-like type</see>.</returns>
+    public static bool IsTaskLikeType(this ITypeSymbol typeSymbol)
```
josefpihrt, line 721: This method is used only once and that's why I would just inline it into `IsAwaitableTaskType` + if it's inlined we know that condition

```cs
if (definition.HasMetadataName(in MetadataNames.System_Threading_Tasks_Task)
    || definition.HasMetadataName(in MetadataNames.System_Threading_Tasks_Task_T))
{
    return true;
}
```

is redundant (it will never succeed).

src/Core/SymbolUtility.cs
```diff
@@ -623,53 +624,66 @@ public static ulong GetEnumValueAsUInt64(object value, INamedTypeSymbol enumType
...
+    {
+        ITypeSymbol? typeSymbol = (symbol as ITypeSymbol) ?? (symbol as IMethodSymbol)?.ReturnType;
 
-        if (namedTypeSymbol is null)
+        if (typeSymbol is null or { SpecialType: SpecialType.System_Void })
             return false;
 
-        INamedTypeSymbol originalDefinition = namedTypeSymbol.OriginalDefinition;
-        TypeKind typeKind = originalDefinition.TypeKind;
+        if (typeSymbol.IsWellKnownTaskType())
```
josefpihrt, line 650: When this method is called from `IsAwaitableTaskType` we know that this check will never succeed.

Could you please add private overload of `IsAwaitable` which does not contain this check and will be called from `IsAwaitableTaskType`.

josefpihrt (changes requested):

josefpihrt (approved):

josefpihrt:
@Govorunb Great PR! Thanks for the contribution.

PR #1593, opened 2024-12-06
Add analyzer RCS0062, put expression body on its own line (#1575)

### LINKED ISSUE #1575: Analyzer that checks the breaking behavior of expression bodies
In our code we prefer having single line methods and properties as expression bodies with 
```
roslynator_body_style = expression
roslynator_use_block_body_when_declaration_spans_over_multiple_lines = true
roslynator_use_block_body_when_expression_spans_over_multiple_lines = true
```
However, we also like to break before `=>` to easier see the actual body.

For that I would like to request an analyzer and fix.

**Example**
```csharp
public int Property
    => 12;

public int DoSomething()
    => SingleLineBreaksBeforeArray();
```

What do you think of it?

It could be an option like
```
roslynator_expression_body_new_line = never|after_arrow|before_arrow
```

cbersch:
Fixes #1575 

Add new analyzer to put an existing expression body on its own line.

### DIFF
--- a/src/Analyzers.xml
+++ b/src/Analyzers.xml
@@ -1995,6 +1995,7 @@ foreach (var item in items)
       <Option Key="body_style" IsRequired="true" />
       <Option Key="use_block_body_when_declaration_spans_over_multiple_lines" IsRequired="true" />
       <Option Key="use_block_body_when_expression_spans_over_multiple_lines" IsRequired="true" />
+      <Option Key="arrow_token_new_line" IsRequired="false" />
     </ConfigOptions>
     <Options>
       <Option Identifier="ConvertExpressionBodyToBlockBodyWhenExpressionIsMultiLine">
--- a/src/Analyzers.xml
+++ b/src/Analyzers.xml
@@ -1630,6 +1630,23 @@ public class C
     </Samples>
   </Analyzer>
   <Analyzer>
+    <Id>RCS0062</Id>
+    <Identifier>PutExpressionBodyOnItsOwnLine</Identifier>
+    <Title>Put expression body on its own line</Title>
+    <DefaultSeverity>Info</DefaultSeverity>
+    <IsEnabledByDefault>false</IsEnabledByDefault>
+    <Samples>
+      <Sample>
+        <Before><![CDATA[object Foo() => null;]]></Before>
+        <After><![CDATA[object Foo()
+    => null;]]></After>
+      </Sample>
+    </Samples>
+    <ConfigOptions>
+      <Option Key="arrow_token_new_line" IsRequired="false" />
+    </ConfigOptions>
+  </Analyzer>
+  <Analyzer>
     <Id>RCS1001</Id>
     <Identifier>AddBracesWhenExpressionSpansOverMultipleLines</Identifier>
     <Title>Add braces (when expression spans over multiple lines)</Title>
--- a/src/Common/CSharp/Analysis/ConvertExpressionBodyAnalysis.cs
+++ b/src/Common/CSharp/Analysis/ConvertExpressionBodyAnalysis.cs
@@ -1,0 +1,24 @@
+﻿using Microsoft.CodeAnalysis.CSharp;
+
+namespace Roslynator.CSharp.Analysis;
+
+internal static class ConvertExpressionBodyAnalysis
+{
+    public static bool AllowPutExpressionBodyOnItsOwnLine(SyntaxKind syntaxKind)
+    {
+        // allow putting expression on new line for all method-like declarations, except for accessors.
+        switch (syntaxKind)
+        {
+            case SyntaxKind.MethodDeclaration:
+            case SyntaxKind.ConstructorDeclaration:
+            case SyntaxKind.DestructorDeclaration:
+            case SyntaxKind.PropertyDeclaration:
+            case SyntaxKind.IndexerDeclaration:
+            case SyntaxKind.OperatorDeclaration:
+            case SyntaxKind.ConversionOperatorDeclaration:
+                return true;
+            default:
+                return false;
+        }
+    }
+}
--- a/src/Tests/Analyzers.Tests/RCS1016UseExpressionBodyTests.cs
+++ b/src/Tests/Analyzers.Tests/RCS1016UseExpressionBodyTests.cs
@@ -12,6 +12,11 @@ public class RCS1016UseExpressionBodyTests : AbstractCSharpDiagnosticVerifier<Us
 {
     public override DiagnosticDescriptor Descriptor { get; } = DiagnosticRules.UseBlockBodyOrExpressionBody;
 
+    public override CSharpTestOptions Options
+    {
+        get { return base.Options.AddConfigOption(ConfigOptionKeys.BodyStyle, ConfigOptionValues.BodyStyle_Expression); }
+    }
+
     [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.UseBlockBodyOrExpressionBody)]
     public async Task Test_Constructor()
     {
@@ -32,7 +37,33 @@ class C
 
     void M() { }
 }
-", options: Options.AddConfigOption(ConfigOptionKeys.BodyStyle, ConfigOptionValues.BodyStyle_Expression));
+", options: Options);
+    }
+
+    [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.UseBlockBodyOrExpressionBody)]
+    public async Task Test_Constructor_BreakBeforeArrow()
+    {
+        await VerifyDiagnosticAndFixAsync(@"
+class C
+{
+    public C()
+    [|{
+        M();
+    }|]
+
+    void M() { }
+}
+", @"
+class C
+{
+    public C()
+        => M();
+
+    void M() { }
+}
+",
+        options: Options.AddConfigOption(ConfigOptionKeys.ArrowTokenNewLine, ConfigOptionValues.ArrowTokenNewLine_Before),
+        additionalDiagnostics: new DiagnosticDescriptor[] { DiagnosticRules.PutExpressionBodyOnItsOwnLine });
     }
 
     [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.UseBlockBodyOrExpressionBody)]
@@ -55,7 +86,7 @@ class C
 
     void M() { }
 }
-", options: Options.AddConfigOption(ConfigOptionKeys.BodyStyle, ConfigOptionValues.BodyStyle_Expression));
+", options: Options);
     }
 
     [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.UseBlockBodyOrExpressionBody)]
@@ -74,7 +105,7 @@ class C
 {
     string M() => null;
 }
-", options: Options.AddConfigOption(ConfigOptionKeys.BodyStyle, ConfigOptionValues.BodyStyle_Expression));
+", options: Options);
     }
 
     [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.UseBlockBodyOrExpressionBody)]
@@ -97,7 +128,7 @@ class C
         x,
         y);
 }
-", options: Options.AddConfigOption(ConfigOptionKeys.BodyStyle, ConfigOptionValues.BodyStyle_Expression));
+", options: Options);
     }
 
     [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.UseBlockBodyOrExpressionBody)]
@@ -122,7 +153,7 @@ class C
             y,
                 z);
 }
-", options: Options.AddConfigOption(ConfigOptionKeys.BodyStyle, ConfigOptionValues.BodyStyle_Expression));
+", options: Options);
     }
 
     [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.UseBlockBodyOrExpressionBody)]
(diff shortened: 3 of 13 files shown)


cbersch:
@dotnet-policy-service agree

src/Tests/Formatting.Analyzers.Tests/RCS0062PutExpressionBodyOnItsOwnLineTests.cs
```diff
@@ -0,0 +1,276 @@
...
+{
+    public override DiagnosticDescriptor Descriptor { get; } = DiagnosticRules.PutExpressionBodyOnItsOwnLine;
+
+    [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.PutExpressionBodyOnItsOwnLine)]
+    public async Task Test_Constructor_BreakAfterArrow()
+    {
+        await VerifyDiagnosticAndFixAsync(@"
+class C
+{
+    public C() =>[||] M();
```
josefpihrt, line 21: If you change diagnostic span to a single space the test are passing.

cbersch:
@josefpihrt So now I tried to add bool option `roslynator_expression_body_style_on_next_line`. 
The current implementation doesn't work, where I add a new line to the arrow token itself depending on the desired options.
I added one test, which currently fails.

I decided to go with the bool option instead of adding a third value for the `expression_body_style`, so that the RR0169 can be configured. Otherwise we would need a third refactoring in addition to RR0038 and RR0169 to "Convert block body to expression body on new line" to cover that.

src/Common/CSharp/Analysis/ConvertExpressionBodyAnalysis.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Roslynator.CSharp.Analysis;
+internal static class ConvertExpressionBodyAnalysis
```
josefpihrt, line 6:
```suggestion
namespace Roslynator.CSharp.Analysis;

internal static class ConvertExpressionBodyAnalysis
```

src/Common/CSharp/Analysis/ConvertExpressionBodyAnalysis.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            case SyntaxKind.MethodDeclaration:
+            case SyntaxKind.ConstructorDeclaration:
+            case SyntaxKind.DestructorDeclaration:
+            case SyntaxKind.PropertyDeclaration:
+            case SyntaxKind.IndexerDeclaration:
+            case SyntaxKind.OperatorDeclaration:
+            case SyntaxKind.ConversionOperatorDeclaration:
+                return true;
+
```
josefpihrt, line 26:
```suggestion
                return true;
```

src/Common/CSharp/Analysis/ConvertExpressionBodyAnalysis.cs
```diff
@@ -0,0 +1,31 @@
...
+            default:
```
josefpihrt, line 27:
```suggestion
                return true;
            default:
```

src/ConfigOptions.xml
```diff
@@ -119,6 +119,10 @@
       <Value>before</Value>
     </Values>
   </Option>
+  <Option Id="ExpressionBodyStyleOnNextLine">
+    <ValuePlaceholder>true|false</ValuePlaceholder>
+    <Description>Place expression body always on next/same line</Description>
+  </Option>
```
josefpihrt, line 125: Could you please explain why did you add this option? I think that this option is basically the same as checking if analyzer is enabled or not. If you want to decide whether to wrap line before `=>` in RCS1016 (as an example) just check if RCS0062 is effective or not.

See extensions methods `IsEffective` for `DiagnosticDescriptor`.

cbersch: Basically you are right, but I wouldn't know how to use that information in the `CodeFixProvider`.
I found usings of `DiagnosticDescriptor.IsEffective` only in analyzers. The respective information is then unambiguous via the created diagnostic.

But configuring the behavior of a single diagnostic (RCS1016) in the `CodeFixProvider` is always done via `AnalyzerConfigOptions`.

josefpihrt: You can definitely get the information whether an analyzer is effective from `CodeFixProvider`. You need a `SyntaxTree`, which is property of a node and `CompilationOptions` which is a property of a `Project`:

```cs
DiagnosticRules.ConfigureAwait.IsEffective(
    node.SyntaxTree,
    context.Document.Project.CompilationOptions,
    cancellationToken);
```

cbersch: Thanks! I'm still making my way through to understand what to get from where.

So that would work, but now I would have dependency problems:
Currently, my attempt for the (not yet working) line breaking is in `Workspaces.Common.ConvertBlockBodyToExpressionBodyRefactoring`, but the `DiagnosticRules` are defined in `Formatting.Analyzers`.

To know, how to refactor that, I would need to know, how the actual new line is inserted. My approach doesn't work, yet.
Probably I would need to call the `SyntaxNode ConvertBlockBodyToExpressionBodyRefactoring.Refactor(SyntaxNode, AnalyzerConfigOptions)` from somewhere else, then manipulated the new syntax node, and finally replace the node in the document.

josefpihrt: Could you call `IsEffective` from code fix provider and then pass it to `ConvertBlockBodyToExpressionBodyRefactoring ` as a parameter? Something like `bool wrapLineBeforeArrowToken`?

cbersch: That also doesn't work, because the RCS0062 DiagnosticRule is defined in `Formatting.Analyzers`, but the caller of `ConvertBlockBodyToExpressionBodyRefactoring.RefactorAsync`, which would then have to evaluate the bool, like `UseBlockBodyOrExpressionBodyCodeFixProvider` (RCS1016) or `ConstructorDeclarationRefactoring` can't reference that project.

josefpihrt: @cbersch I moved `DiagnosticRules` and `DiagnosticIdentifiers` to `Roslynator.Common` so now it's accessible from everywhere. Please re-run `generate_code.ps1`.

cbersch: Grest, thanks. I'll rework when I'm back from vacaton

cbersch: @josefpihrt Now the `ConvertBlockBodyToExpressionBodyRefactoring` respects the `RCS0062` diagnostic in order to determine if to wrap on a new line or not.
I added only a single test for RCS1016 and RR0169, because the line wrapping itself doesn't work, yet, for these. There I would again need your help.

src/Tests/Analyzers.Tests/RCS1016UseExpressionBodyTests.cs
```diff
@@ -32,7 +37,33 @@ class C
...
+class C
+{
+    public C()
+        => M();
+
+    void M() { }
+}
+",
+        options: Options.AddConfigOption(ConfigOptionKeys.ArrowTokenNewLine, ConfigOptionValues.ArrowTokenNewLine_Before),
+        additionalDiagnostics: new DiagnosticDescriptor[] { DiagnosticRules.PutExpressionBodyOnItsOwnLine });
```
josefpihrt, line 66: This is not necessary. You can use `Options.EnableDiagnostic(DiagnosticRules.PutExpressionBodyOnItsOwnLine)`

src/Common/CSharp/Analysis/ConvertExpressionBodyAnalysis.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.CodeAnalysis.CSharp;
+
+namespace Roslynator.CSharp.Analysis;
+
+internal static class ConvertExpressionBodyAnalysis
+{
+    public static bool BreakExpressionOnNewLine(SyntaxKind syntaxKind)
```
josefpihrt, line 7: What is the purpose of this function. To me it seems that you want to exclude accessors.

src/ConfigOptions.xml
josefpihrt: @cbersch  I tried to fix tests. You can find it in this commit: https://github.com/dotnet/roslynator/commit/2ac2be6f236a9596387f4949f31af8e9b770ee9e

src/Common/CSharp/Analysis/ConvertExpressionBodyAnalysis.cs
cbersch: Yes, accessors should not be affected. Should I add a comment, oder change the method name?

src/ConfigOptions.xml
cbersch: @josefpihrt Thanks, I cherry-picked that commit to my branch. Looks good.

src/Tests/Analyzers.Tests/RCS1016UseExpressionBodyTests.cs
cbersch: Thanks, using that one and reverted the `additionalDiagnostics` parameter changes.

src/Workspaces.Common/CSharp/Refactorings/ConvertBlockBodyToExpressionBodyRefactoring.cs
```diff
@@ -179,19 +181,43 @@ public static Task<Document> RefactorAsync(
...
         CancellationToken cancellationToken = default)
     {
         AnalyzerConfigOptions configOptions = document.GetConfigOptions(node.SyntaxTree);
+        NewLinePosition newLinePosition = GetNewLinePosition(document, node, configOptions, cancellationToken);
+
+        SyntaxNode newNode = Refactor(node, configOptions, newLinePosition).WithFormatterAnnotation();
+
+        if (newLinePosition == NewLinePosition.After)
+        {
+            var arrowToken = CSharpUtility.GetExpressionBody(newNode).ArrowToken;
```
josefpihrt, line 196:
```suggestion
            ArrowExpressionClauseSyntax arrowToken = CSharpUtility.GetExpressionBody(newNode).ArrowToken;
```

src/Workspaces.Common/CSharp/Refactorings/ConvertBlockBodyToExpressionBodyRefactoring.cs
```diff
@@ -179,19 +181,43 @@ public static Task<Document> RefactorAsync(
...
         CancellationToken cancellationToken = default)
     {
         AnalyzerConfigOptions configOptions = document.GetConfigOptions(node.SyntaxTree);
+        NewLinePosition newLinePosition = GetNewLinePosition(document, node, configOptions, cancellationToken);
+
+        SyntaxNode newNode = Refactor(node, configOptions, newLinePosition).WithFormatterAnnotation();
+
+        if (newLinePosition == NewLinePosition.After)
+        {
+            ArrowExpressionClauseSyntax arrowToken = CSharpUtility.GetExpressionBody(newNode).ArrowToken;
```
josefpihrt, line 196:
```suggestion
            SyntaxToken arrowToken = CSharpUtility.GetExpressionBody(newNode).ArrowToken;
```

cbersch:
@josefpihrt Thanks for the fixes, hopefully I'll get into the coding conventions quickly.
I couldn't reach the last unresolved conversation anymore after my last force push.

josefpihrt:
@cbersch please update changelog

cbersch:
@josefpihrt Added to changelog

src/Common/CSharp/Analysis/ConvertExpressionBodyAnalysis.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Microsoft.CodeAnalysis.CSharp;
+
+namespace Roslynator.CSharp.Analysis;
+
+internal static class ConvertExpressionBodyAnalysis
+{
+    public static bool AllowPutExpressionBodyOnItsOwnLine(SyntaxKind syntaxKind)
+    {
+        // allow putting expression on new line for all method-like declarations, except for accessors.
+        switch (syntaxKind)
```
josefpihrt, line 10: What about local function?

src/Analyzers.xml
```diff
@@ -1629,6 +1629,23 @@ public class C
       </Sample>
     </Samples>
   </Analyzer>
+  <Analyzer>
+    <Id>RCS0062</Id>
```
josefpihrt, line 1633: Could you please add `<Option Key="arrow_token_new_line" IsRequired="false" />` to RCS1016 and RR0169?

src/Tests/Analyzers.Tests/RCS1016UseExpressionBodyTests.cs
```diff
@@ -12,6 +12,11 @@ public class RCS1016UseExpressionBodyTests : AbstractCSharpDiagnosticVerifier<Us
 {
     public override DiagnosticDescriptor Descriptor { get; } = DiagnosticRules.UseBlockBodyOrExpressionBody;
 
+    public override CSharpTestOptions Options
+    {
+        get { return base.Options.AddConfigOption(ConfigOptionKeys.BodyStyle, ConfigOptionValues.BodyStyle_Expression); }
```
josefpihrt, line 17: 👍 

Could you please store it in a field?

josefpihrt (commented):
Looking really good. I added some final comments and then we are good to go.

josefpihrt (approved):

josefpihrt:
@cbersch  Great job! I think you were one of the first (if not the first) contributor to implement a new analyzer (including the code fix). Thanks!

cbersch:
@josefpihrt Thanks for your support, I'm happy to contribute.
Kudos for your awesome tool!

PR #1609, opened 2025-01-19
Support custom path of a test file

josefpihrt:


src/Tests/Testing.Common/Testing/CodeVerifier.cs
```diff
@@ -362,12 +362,12 @@ internal static (Document document, ImmutableArray<ExpectedDocument> expectedDoc
...
             for (int i = 0; i < additionalFiles.Length; i++)
             {
-                string documentName = AppendNumberToFileName(options.DocumentName, i + 2);
+                string documentName = Path.GetFileName(additionalFiles[i].Path) ?? AppendNumberToFileName(options.DocumentName, i + 2);
 
                 Document additionalDocument = project.AddDocument(
                     documentName,
                     SourceText.From(additionalFiles[i].Source),
-                    filePath: Path.Combine(directoryPath, documentName));
+                    filePath: Path.Combine(directoryPath, additionalFiles[i].Path ?? documentName));
```
norritt, line 370: I'd suggest
```suggestion
                    filePath: Path.Combine(directoryPath, Path.GetDirectoryName(additionalFiles[i].Path ?? documentName), documentName));
```
that should properly process passed paths without file name, like `Foo/Bar/` or something like this

src/Tests/Testing.Common/Testing/CodeVerifier.cs
```diff
@@ -362,12 +362,12 @@ internal static (Document document, ImmutableArray<ExpectedDocument> expectedDoc
 
             for (int i = 0; i < additionalFiles.Length; i++)
             {
-                string documentName = AppendNumberToFileName(options.DocumentName, i + 2);
+                string documentName = Path.GetFileName(additionalFiles[i].Path) ?? AppendNumberToFileName(options.DocumentName, i + 2);
```
norritt, line 365: in case I am providing only a path, it could work erroneously. I do not know how it would behave for cases `Foo/Bar` and `Foo/Bar/`

src/Tests/Testing.Common/Testing/CodeVerifier.cs
josefpihrt: I added validation of file path

src/Tests/Testing.Common/Testing/CodeVerifier.cs
josefpihrt: I added validation of file path

PR #1648, opened 2025-05-01
Update Analyzers.xml

BenjaminBrienen:
remove removed analyzer from Analyzers.xml

Sorry that the diff algorithm makes the change look weird


josefpihrt:
Why did you remove this particular analyzer?

BenjaminBrienen:
Ok. I couldn't even find a reference to it in the repo other than the changelog, but I understand if you want to keep it and close this.

PR #1661, opened 2025-06-20
Change Span of PutFullAccessorOnOwnLine to be correct

niekschoemaker:
The span used to be empty causing the warning to only appear when building, and not in the IDE (at least for Rider) unless something like Error lens is active.

I changed it to match the span it expects for the code fix, and also removed the break because both accessors can be wrong, and this way it correctly marks both of them as being on the wrong line. Otherwise applying the code fix would suddenly cause a new warning for the other accessor.


josefpihrt:
And sorry for late review 😊

josefpihrt (approved):

josefpihrt:
@niekschoemaker Thanks for the contribution!

PR #1720, opened 2025-12-06
Fix CLI rename-symbol scope option not being applied correctly

andrtmschkw:


josefpihrt (dismissed):

josefpihrt:
@andrtmschkw Please update changelog, otherwise fine.

josefpihrt (approved):

PR #1721, opened 2025-12-06
Fix CLI rename-symbol support for top-level statement

andrtmschkw:
For Program.cs with top-level staments:
`
System.Console.WriteLine(args.Length);
var y = 99;
class X {}
interface I {}
`

rename-symbol fails in because ISymbol for args is ImplicitlyDeclared=true and the Locations are empty but it is accesses without a bounds check.

`System.IndexOutOfRangeException: Index was outside the bounds of the array.
at Roslynator.Rename.SymbolRenameState.<>c__DisplayClass30_0.<AnalyzeProjectAsync>b__0(ISymbol symbol) in \src\Workspaces.Core\Rename\SymbolRenameState.cs:line 173`

Also LocalSymbolFinder.FindLocalSymbols does not support nodes of SyntaxKind.CompilationUnit (top-level statement) and the assertion fails.

### DIFF
--- a/src/CommandLine/Commands/RenameSymbolCommand.cs
+++ b/src/CommandLine/Commands/RenameSymbolCommand.cs
@@ -86,9 +86,9 @@ internal class RenameSymbolCommand : MSBuildWorkspaceCommand<RenameSymbolCommand
         {
             var options = new SymbolRenamerOptions()
             {
-                SkipTypes = (ScopeFilter & RenameScopeFilter.Type) != 0,
-                SkipMembers = (ScopeFilter & RenameScopeFilter.Member) != 0,
-                SkipLocals = (ScopeFilter & RenameScopeFilter.Local) != 0,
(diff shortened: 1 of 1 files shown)


andrtmschkw:
@dotnet-policy-service agree

src/Workspaces.Core/Rename/SymbolListHelpers.cs
```diff
@@ -93,7 +93,7 @@ public static List<ISymbol> SortAndFilterMemberSymbols(IEnumerable<ISymbol> symb
                     results.Add(symbol);
 
                 results.AddRange(methodSymbol.TypeParameters);
-                results.AddRange(methodSymbol.Parameters);
+                results.AddRange(methodSymbol.Parameters.Where(ms => !ms.IsImplicitlyDeclared));
```
josefpihrt:
```suggestion
                results.AddRange(methodSymbol.Parameters.Where(p => !p.IsImplicitlyDeclared));
```

PR #1742, opened 2026-02-03
Fix RCS1077 formatting when converting Select().ToList() to ConvertAll()

### LINKED ISSUE #1728: [RCS1077] Rule produces incorrect formatting when applying code fixes (; placed on wrong indentation level)
### Product and Version Used
Roslynator Version: 4.14.1
IDE: Visual Studio 2022
Framework: .NET 8.0

### Description
When running dotnet format or using the Roslyn CLI with Roslynator enabled, the code fix for RCS1077 incorrectly formats the generated code. Specifically, the semicolon after a transformed method call is placed at an incorrect indentation level.

### Steps to Reproduce
- Enable rule RCS1077.
- Write a LINQ query with formatting issues.
- Call dotnet format or Roslynator CLI to fix issues.

**Input**
```
eventPublisher.EnqueueLowPriorityEvent(() =>
{
    IReadOnlyList<IEvent> events = normalizedSpecialistBios
        .Select(sb => new SpecialistBioUpdatedEvent(sb.ProfileId))
        .ToList();

    return new BatchEvent(events);
});
```

**Actual Behavior**

```
eventPublisher.EnqueueLowPriorityEvent(() =>
{
    IReadOnlyList<IEvent> events = normalizedSpecialistBios
        .ConvertAll(sb => new SpecialistBioUpdatedEvent(sb.ProfileId))
;
    return new BatchEvent(events);
});
```
Notice that the semicolon is placed on a new line with indentation based on the start of the original expression rather than aligned with the transformed method call.
It appears that the indentation is calculated relative to the original Select(...).ToList() call instead of the new ConvertAll(...) call inserted by the code fix.

**Expected Behavior**
The semicolon should remain properly aligned with the preceding method call, e.g.:
```
IReadOnlyList<IEvent> events = normalizedSpecialistBios
    .ConvertAll(sb => new SpecialistBioUpdatedEvent(sb.ProfileId));
```

MattFromRVA:
Fixes #1728

### DIFF
--- a/src/Analyzers.CodeFixes/CSharp/CodeFixes/OptimizeLinqMethodCallCodeFixProvider.cs
+++ b/src/Analyzers.CodeFixes/CSharp/CodeFixes/OptimizeLinqMethodCallCodeFixProvider.cs
@@ -587,11 +587,26 @@ public sealed class OptimizeLinqMethodCallCodeFixProvider : BaseCodeFixProvider
         in SimpleMemberInvocationExpressionInfo invocationInfo,
         CancellationToken cancellationToken)
     {
-        InvocationExpressionSyntax invocationExpression2 = SimpleMemberInvocationExpressionInfo(invocationInfo.Expression).InvocationExpression;
+        InvocationExpressionSyntax toListInvocation = invocationInfo.InvocationExpression;
 
-        InvocationExpressionSyntax newInvocationExpression = ChangeInvokedMethodName(invocationExpression2, "ConvertAll");
+        InvocationExpressionSyntax selectInvocation = SimpleMemberInvocationExpressionInfo(invocationInfo.Expression).InvocationExpression;
 
-        return document.ReplaceNodeAsync(invocationInfo.InvocationExpression, newInvocationExpression, cancellationToken);
+        InvocationExpressionSyntax newInvocationExpression = ChangeInvokedMethodName(selectInvocation, "ConvertAll");
+
+        IEnumerable<SyntaxTrivia> removedTrivia = toListInvocation.DescendantTrivia(
+            TextSpan.FromBounds(selectInvocation.Span.End, toListInvocation.Span.End));
+
+        if (removedTrivia.Any(f => !f.IsWhitespaceOrEndOfLineTrivia()))
+        {
+            newInvocationExpression = newInvocationExpression.WithTrailingTrivia(
+                removedTrivia.Where(f => !f.IsWhitespaceOrEndOfLineTrivia()).Concat(toListInvocation.GetTrailingTrivia()));
+        }
+        else
+        {
+            newInvocationExpression = newInvocationExpression.WithTrailingTrivia(toListInvocation.GetTrailingTrivia());
+        }
+
+        return document.ReplaceNodeAsync(toListInvocation, newInvocationExpression, cancellationToken);
     }
 
     private static Task<Document> CallSumInsteadOfSelectManyAndCountAsync(
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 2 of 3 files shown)


josefpihrt (commented):
Nice fix — the root cause (the new node inheriting the `Select` invocation's trailing trivia, which carried the newline before `.ToList()`) is correctly addressed by anchoring to `toListInvocation.GetTrailingTrivia()`, and it stays backward-compatible with the single-line case. A couple of inline comments below; one I'd suggest resolving before merge.

**Test coverage:** could we add a case for a single-line `//` comment in the removed segment (see the inline comment on the `if`), and optionally one with an `IReadOnlyList<...>` target to mirror the exact repro in #1728? The current tests only cover `/* */`, which is what lets the `//` case slip through.

src/Analyzers.CodeFixes/CSharp/CodeFixes/OptimizeLinqMethodCallCodeFixProvider.cs
```diff
@@ -587,11 +587,26 @@ private static Task<Document> CallConvertAllInsteadOfSelectAsync(
...
+        InvocationExpressionSyntax selectInvocation = SimpleMemberInvocationExpressionInfo(invocationInfo.Expression).InvocationExpression;
 
-        return document.ReplaceNodeAsync(invocationInfo.InvocationExpression, newInvocationExpression, cancellationToken);
+        InvocationExpressionSyntax newInvocationExpression = ChangeInvokedMethodName(selectInvocation, "ConvertAll");
+
+        IEnumerable<SyntaxTrivia> removedTrivia = toListInvocation.DescendantTrivia(
+            TextSpan.FromBounds(selectInvocation.Span.End, toListInvocation.Span.End));
+
+        if (removedTrivia.Any(f => !f.IsWhitespaceOrEndOfLineTrivia()))
+        {
```
josefpihrt, line 600: This branch strips all whitespace/EOL trivia and reattaches surviving comments inline. That's fine for block comments, but a single-line `//` comment has no terminating newline of its own — the newline is separate `EndOfLineTrivia`, which gets stripped here. So this input:

```csharp
var x = items
    .Select(f => f.ToString()) // keep this
    .ToList();
```

would become:

```csharp
var x = items
    .ConvertAll(f => f.ToString())// keep this;
```

…where the `;` is now commented out → compile error. Could we preserve a trailing newline when the salvaged trivia ends in a single-line comment (or otherwise guard against it)? A `//`-comment variant of `Test_CallConvertAllInsteadOfSelectAndToList_PreservesCommentInRemovedSegment` would reproduce it.

src/Analyzers.CodeFixes/CSharp/CodeFixes/OptimizeLinqMethodCallCodeFixProvider.cs
```diff
@@ -587,11 +587,26 @@ private static Task<Document> CallConvertAllInsteadOfSelectAsync(
...
+            newInvocationExpression = newInvocationExpression.WithTrailingTrivia(
+                removedTrivia.Where(f => !f.IsWhitespaceOrEndOfLineTrivia()).Concat(toListInvocation.GetTrailingTrivia()));
+        }
+        else
+        {
```
josefpihrt, line 605: Nit: both branches converge on `... .Concat(toListInvocation.GetTrailingTrivia())` — when there are no comments the `Where` yields nothing, so this collapses to a single path:

```csharp
IEnumerable<SyntaxTrivia> comments = removedTrivia.Where(f => !f.IsWhitespaceOrEndOfLineTrivia());
newInvocationExpression = newInvocationExpression.WithTrailingTrivia(
    comments.Concat(toListInvocation.GetTrailingTrivia()));
```

Behavior-equivalent and drops the branch. Also worth materializing `removedTrivia` with `.ToList()` since it's currently enumerated multiple times.

josefpihrt (changes requested):

PR #1750, opened 2026-03-03
Fix RCS1166 to report diagnostic when null is on the left side of comparison

### LINKED ISSUE #1698: RCS1166 doesn't work if first operand is null
The following code should both raise an `RCS1166` diagnostic, but the `null == a` isn't recognized. Tested with 4.14.1

```csharp
public static class Class1
{
    public static bool Correct(ImmutableArray<int> a)
    {
        return a == null;
    }

    public static bool Wrong(ImmutableArray<int> a)
    {
        return null == a;
    }
}
```

Image

Until now I couldn't reproduce this with any other custom type, which wouldn't also trigger CS8073.

MattFromRVA:
Fixes #1698


josefpihrt (approved):

josefpihrt:
@MattFromRVA Thank your for your contribution!

PR #1758, opened 2026-04-11
[CLI] Fix fix command ignoring --include / --exclude file filter

### LINKED ISSUE #1757: [CLI] fix command ignores --include / --exclude file filter
The `--include` and `--exclude` options on the `fix` command do not filter which files are fixed. All files in the solution are modified regardless of the specified patterns.

The `analyze` command correctly respects these options (fixed in #1459).

Reproduction: https://github.com/hashiiiii/roslynator-fix-include-bug-repro

```bash
roslynator fix ./TestSolution.sln \
  --include "ProjectA/**" \
  --supported-diagnostics CA1822
```

**Expected:** Only files under `ProjectA/` are fixed.
**Actual:** Files under `ProjectB/` are also fixed.

In `CodeFixer.cs`, `FileSystemFilter` is not checked in `GetFixableDiagnostics` or `FixDiagnosticsAsync`. It is only checked in `GetDiagnosticsAsync`, which runs after the fix loop for reporting purposes.

hashiiiii:
fixes #1757

### DIFF
--- a/src/Workspaces.Core/CodeFixes/CodeFixer.cs
+++ b/src/Workspaces.Core/CodeFixes/CodeFixer.cs
@@ -294,12 +294,16 @@ internal class CodeFixer
             IEnumerable<Diagnostic> fixableCompilerDiagnostics = compilerDiagnostics
                 .Where(f => f.Severity != DiagnosticSeverity.Error
                     && !Options.IgnoredCompilerDiagnosticIds.Contains(f.Id)
-                    && fixersById.ContainsKey(f.Id));
+                    && fixersById.ContainsKey(f.Id)
+                    && (f.Location.SourceTree is null
+                        || Options.FileSystemFilter?.IsMatch(f.Location.SourceTree.FilePath) != false));
 
             return diagnostics
                 .Where(f => f.IsEffective(Options, project.CompilationOptions)
                     && analyzersById.ContainsKey(f.Id)
-                    && fixersById.ContainsKey(f.Id))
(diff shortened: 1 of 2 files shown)


hashiiiii:
@dotnet-policy-service agree

src/Workspaces.Core/CodeFixes/CodeFixer.cs
```diff
@@ -294,12 +294,16 @@ ImmutableArray<Diagnostic> GetFixableDiagnostics(
...
+                    && (f.Location.SourceTree is null
+                        || Options.FileSystemFilter?.IsMatch(f.Location.SourceTree.FilePath) != false));
 
             return diagnostics
                 .Where(f => f.IsEffective(Options, project.CompilationOptions)
                     && analyzersById.ContainsKey(f.Id)
-                    && fixersById.ContainsKey(f.Id))
+                    && fixersById.ContainsKey(f.Id)
+                    && (f.Location.SourceTree is null
+                        || Options.FileSystemFilter?.IsMatch(f.Location.SourceTree.FilePath) != false))
```
hashiiiii, line 306: Referenced the existing pattern.
https://github.com/dotnet/roslynator/blob/main/src/Workspaces.Core/Diagnostics/CodeAnalyzer.cs#L214-L215

josefpihrt (approved):

josefpihrt:
@hashiiiii Thank you for your contribution!

Issue #1748, opened 2026-02-23
CLI does not work in Ubuntu 24.04 and .NET 10 SDK


jdvor:
Roslynator CLI does not work on Ubuntu 24.04.4 LTS - both locally and in all GitHub CI pipelines based on Linux OS. 

The same tool version and .NET SDK, tested with the same project, works on Windows. This is an OS-specific issue. 
Given the importance of Linux as the OS for most CI pipelines (certainly on GitHub, which runs on some version of Ubuntu), this is not a fringe environment. 

It might be related to .NET 10 SDK, as we have used `roslynator` successfully in the past with .NET 8.0 and .NET 9.0 SDKs in an Ubuntu environment.

## command
```
❯ roslynator analyze "./terraverse-vterm.sln" --exclude "./tests/**/*.csproj" -o "./artifacts/code-quality/roslynator.xml" --severity-level "warning" -v normal
Loading solution '/home/jand/dev/terraverse-vterm/terraverse-vterm.sln'...
System.AggregateException: One or more errors occurred. (Field not found: 'Microsoft.Build.Shared.MSBuildConstants.InvalidPathChars'.)
 ---> System.MissingFieldException: Field not found: 'Microsoft.Build.Shared.MSBuildConstants.InvalidPathChars'.
   at Microsoft.Build.Construction.SolutionFile.ValidateProjectRelativePath(ProjectInSolution proj)
   at Microsoft.Build.Construction.SolutionFile.ParseFirstProjectLine(String firstLine, ProjectInSolution proj)
   at Microsoft.Build.Construction.SolutionFile.ParseProject(String firstLine)
   at Microsoft.Build.Construction.SolutionFile.ParseSolution()
   at Microsoft.Build.Construction.SolutionFile.ParseSolutionFile()
   at Microsoft.Build.Construction.SolutionFile.Parse(String solutionFile)
   at Microsoft.CodeAnalysis.MSBuild.MSBuildProjectLoader.LoadSolutionInfoAsync(String solutionFilePath, IProgress`1 progress, ILogger msbuildLogger, CancellationToken cancellationToken)
   at Microsoft.CodeAnalysis.MSBuild.MSBuildWorkspace.OpenSolutionAsync(String solutionFilePath, ILogger msbuildLogger, IProgress`1 progress, CancellationToken cancellationToken)
[...]

Issue #1716, opened 2025-11-30
Roslynator CLI not working on container image of sdk 10.0


gbschenkel:
I am trying to run the `roslynator.dotnet.cli` inside the `mcr.microsoft.com/dotnet/sdk:10.0` I got this issue:
```
root@0c1dd6bd404b:/observability# dotnet tool run roslynator analyze --severity-level info --output roslynator-report.xml --output-format xml
Loading solution '/observability/observability.slnx'...
System.AggregateException: One or more errors occurred. (The project file could not be loaded. System.MissingFieldException: Field not found: 'Microsoft.Build.Shared.MSBuildConstants.InvalidPathChars'.
   at Microsoft.Build.Construction.SolutionFile.ValidateProjectRelativePath(ProjectInSolution proj)
   at Microsoft.Build.Construction.SolutionFile.ReadProjects(SolutionModel solutionModel)
   at Microsoft.Build.Construction.SolutionFile.ReadSolutionModel(SolutionModel solutionModel)
   at Microsoft.Build.Construction.SolutionFile.ParseUsingNewParser()  /observability/observability.slnx)
 ---> Microsoft.Build.Exceptions.InvalidProjectFileException: The project file could not be loaded. System.MissingFieldException: Field not found: 'Microsoft.Build.Shared.MSBuildConstants.InvalidPathChars'.
   at Microsoft.Build.Construction.SolutionFile.ValidateProjectRelativePath(ProjectInSolution proj)
   at Microsoft.Build.Construction.SolutionFile.ReadProjects(SolutionModel solutionModel)
   at Microsoft.Build.Construction.SolutionFile.ReadSolutionModel(SolutionModel solutionModel)
   at Microsoft.Build.Construction.SolutionFile.ParseUsingNewParser()  /observability/observability.slnx
   at Microsoft.Build.Shared.ProjectFileErrorUtilities.VerifyThrowInvalidProjectFile(Boolean condition, String errorSubCategoryResourceName, BuildEventFileInfo projectFile, Exception innerException, String resourceName, Object[] args)
   at Microsoft.Build.Construction.SolutionFile.ParseUsingNewParser()
   at Microsoft.Build.Construction.SolutionFile.Parse(String solutionFile)
   at Microsoft.CodeAnalysis.MSBuild.MSBuildProjectLoader.LoadSolutionInfoAsync(String solutionFilePath, IProgress`1 progress, ILogger msbuildLogger, CancellationToken cancellationToken)
[...]

PR #1762, opened 2026-05-12
[CLI] Fix .slnx parsing on .NET 10 SDK

### LINKED ISSUE #1729: CLI tool give MissingFieldException on Console app template
# roslynator.dotnet.cli 0.12

When I run "Roslynator analyze" from the command line on a new solution based on the Console app template I get `MissingFieldException`. This works on Roslynator version 0.11 

```
.\tools\roslynator.exe analyze C:\Users\X\source\repos\MissingInvalidPathChars\MissingInvalidPathChars.slnx -v diag
Roslynator Version: 0.12.0.0
Roslynator Target Framework: .NET Core
Roslyn Version: 4.14.0.0
Available MSBuild locations:
  .NET Core SDK, Version: 5.0.408, Path: C:\Program Files\dotnet\sdk\5.0.408
  .NET Core SDK, Version: 6.0.428, Path: C:\Program Files\dotnet\sdk\6.0.428
  .NET Core SDK, Version: 7.0.410, Path: C:\Program Files\dotnet\sdk\7.0.410
  .NET Core SDK, Version: 8.0.416, Path: C:\Program Files\dotnet\sdk\8.0.416
  .NET Core SDK, Version: 9.0.307, Path: C:\Program Files\dotnet\sdk\9.0.307
  .NET Core SDK, Version: 10.0.101, Path: C:\Program Files\dotnet\sdk\10.0.101
MSBuild location is 'C:\Program Files\dotnet\sdk\10.0.101'
Loading solution 'C:\Users\X\source\repos\MissingInvalidPathChars\MissingInvalidPathChars.slnx'...
System.AggregateException: One or more errors occurred. (The project file could not be loaded. System.MissingFieldException: Field not found: 'Microsoft.Build.Shared.MSBuildConstants.InvalidPathChars'.
   at Microsoft.Build.Construction.SolutionFile.ValidateProjectRelativePath(ProjectInSolution proj)
   at Microsoft.Build.Construction.SolutionFile.ReadProjects(SolutionModel solutionModel)
   at Microsoft.Build.Construction.SolutionFile.ReadSolutionModel(SolutionModel solutionModel)
   at Microsoft.Build.Construction.SolutionFile.ParseUsingNewParser()  C:\Users\X\source\repos\MissingInvalidPathChars\MissingInvalidPathChars.slnx)
 ---> Microsoft.Build.Exceptions.InvalidProjectFileException: The project file could not be loaded. System.MissingFieldException: Field not found: 'Microsoft.Build.Shared.MSBuildConstants.InvalidPathChars'.
   at Microsoft.Build.Construction.SolutionFile.ValidateProjectRelativePath(ProjectInSolution proj)
   at Microsoft.Build.Construction.SolutionFile.ReadProjects(SolutionModel solutionModel)
   at Microsoft.Build.Construction.SolutionFile.ReadSolutionModel(SolutionModel solutionModel)
   at Microsoft.Build.Construction.SolutionFile.ParseUsingNewParser()  C:\Users\X\source\repos\MissingInvalidPathChars\MissingInvalidPathChars.slnx
   at Microsoft.Build.Shared.ProjectFileErrorUtilities.VerifyThrowInvalidProjectFile(Boolean condition, String errorSubCategoryResourceName, BuildEventFileInfo projectFile, Exception innerException, String resourceName, Object[] args)
   at Microsoft.Build.Shared.ProjectFileErrorUtilities.ThrowInvalidProjectFile(BuildEventFileInfo projectFile, String resourceName, Object[] args)
   at Microsoft.Build.Construction.SolutionFile.ParseUsingNewParser()
   at Microsoft.Build.Construction.SolutionFile.Parse(String solutionFile)
   at Microsoft.CodeAnalysis.MSBuild.MSBuildProjectLoader.LoadSolutionInfoAsync(String solutionFilePath, IProgress`1 progress, ILogger msbuildLogger, CancellationToken cancellationToken)
   at Microsoft.CodeAnalysis.MSBuild.MSBuildWorkspace.OpenSolutionAsync(String solutionFilePath, ILogger msbuildLogger, IProgress`1 progress, CancellationToken cancellationToken)
   at Roslynator.CommandLine.MSBuildWorkspaceCommand`1.OpenProjectOrSolutionAsync(String path, MSBuildWorkspace workspace, IProgress`1 progress, CancellationToken cancellationToken) in /_/src/CommandLine/Commands/MSBuildWorkspaceCommand.cs:line 242
   at Roslynator.CommandLine.MSBuildWorkspaceCommand`1.ExecuteAsync(String path, MSBuildWorkspace workspace, CancellationToken cancellationToken) in /_/src/CommandLine/Commands/MSBuildWorkspaceCommand.cs:line 151
   at Roslynator.CommandLine.MSBuildWorkspaceCommand`1.ExecuteAsync(IEnumerable`1 paths, String msbuildPath, IEnumerable`1 properties) in /_/src/CommandLine/Commands/MSBuildWorkspaceCommand.cs:line 89
   at Roslynator.CommandLine.Program.AnalyzeAsync(AnalyzeCommandLineOptions options) in /_/src/CommandLine/Program.cs:line 346
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at Roslynator.CommandLine.Program.<>c.<Main>b__0_3(MSBuildCommandLineOptions options) in /_/src/CommandLine/Program.cs:line 175
   at CommandLine.ParserResultExtensions.MapResult[T1,T2,TResult](ParserResult`1 result, Func`2 parsedFunc1, Func`2 parsedFunc2, Func`2 notParsedFunc)
   at Roslynator.CommandLine.Program.Main(String[] args) in /_/src/CommandLine/Program.cs:line 169
```

What else can I do to give more context?

charles8051:
## Summary

Fix `.slnx` solution parsing for users running on the .NET 10 SDK, which currently crashes with `TypeLoadException: Could not load type 'Microsoft.Build.Framework.FileUtilities'`.

Fixes #1729, #1748, #1716.

## Root cause

The bundled `Microsoft.Build.Framework.dll` from NuGet is a ref-only assembly stripped of `internal` types. When MSBuild 18 (.NET 10 SDK) parses a `.slnx` solution, `Microsoft.Build.dll` calls the internal `Microsoft.Build.Framework.FileUtilities` type — which only exists in the SDK's full implementation, not in the ref assembly the tool ships. .NET's assembly resolution prefers the locally-bundled ref copy over the SDK's full one, so the type lookup fails.

This was latent on prior SDKs: MSBuild 17's `SolutionFile.Parse` codepath didn't call across into Framework internals, so the wrong-shadowing-of-the-ref-assembly never triggered. SDK 10's refactored slnx parser exposed it.

## Fix

Add explicit `<PackageReference Include="Microsoft.Build.Framework" ExcludeAssets="runtime" />` for net9 and net10, which stops the package from depositing its `.dll` into the tool output. `MSBuildLocator` then loads the SDK's full implementation at runtime, as intended.

Also split the existing `Microsoft.Build` reference by TFM, since `Microsoft.Build` 18.x targets only net10 (and net48) — net9 keeps 17.14.28, net10 moves to 18.4.0.

net8 is intentionally not touched: `.slnx` requires SDK 9.0.200+, so net8 users can't reach the failing codepath. If parity is desired, the same pattern can extend to net8 in a follow-up.

## Test plan

- [x] `dotnet build -p:RoslynatorDotNetCli=true` on net8/net9/net10 — all succeed; net9/net10 outputs no longer contain `Microsoft.Build.Framework.dll`
- [x] `dotnet pack` produces a tool package, locally installed
- [x] `roslynator list-symbols Periphery.slnx --depth member --visibility public` — succeeds on a `.slnx` solution under SDK 10 with no `--msbuild-path` pin (previously crashed with `TypeLoadException`)
- [ ] Maintainer regression check across other CLI commands (`analyze`, `fix`, `format`, etc.) — should be unaffected since the fix is purely about which `.dll` is bundled, not about runtime API usage

### DIFF
--- a/src/CommandLine/CommandLine.csproj
+++ b/src/CommandLine/CommandLine.csproj
@@ -58,7 +58,15 @@
 
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.8.0" />
-    <PackageReference Include="Microsoft.Build" Version="17.14.28" ExcludeAssets="runtime" Condition="'$(TargetFramework)' == 'net9.0' OR '$(TargetFramework)' == 'net10.0'" />
+    <PackageReference Include="Microsoft.Build" Version="17.14.28" ExcludeAssets="runtime" Condition="'$(TargetFramework)' == 'net9.0'" />
+    <PackageReference Include="Microsoft.Build" Version="18.4.0" ExcludeAssets="runtime" Condition="'$(TargetFramework)' == 'net10.0'" />
+    <!-- Suppress bundling of Microsoft.Build.Framework.dll: the NuGet copy is a
+         ref-only assembly that lacks internal types (e.g. FileUtilities) which
+         the SDK's full Microsoft.Build.dll calls when parsing .slnx solutions.
+         MSBuildLocator will load the matching full implementation from the SDK
+         at runtime. See https://github.com/dotnet/roslynator/issues/1729. -->
+    <PackageReference Include="Microsoft.Build.Framework" Version="17.14.28" ExcludeAssets="runtime" Condition="'$(TargetFramework)' == 'net9.0'" />
+    <PackageReference Include="Microsoft.Build.Framework" Version="18.4.0" ExcludeAssets="runtime" Condition="'$(TargetFramework)' == 'net10.0'" />
     <PackageReference Include="Microsoft.Build.Locator" Version="$(RoslynatorMicrosoftBuildLocatorVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="$(RoslynatorCliRoslynVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Workspaces" Version="$(RoslynatorCliRoslynVersion)" />
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- [CLI] Fix `.slnx` parsing on .NET 10 SDK by suppressing the bundled ref-only `Microsoft.Build.Framework.dll`, so MSBuildLocator loads the SDK's full implementation at runtime (fixes [#1729](https://github.com/dotnet/roslynator/issues/1729), [#1748](https://github.com/dotnet/roslynator/issues/1748), [#1716](https://github.com/dotnet/roslynator/issues/1716))
+
 ## [4.15.0] - 2025-12-14
 
 ### Added


charles8051:
@dotnet-policy-service agree

fgimian:
@josefpihrt Any chance this can be considered please? The roslynator CLI is sadly broken on .NET 10 due to this issue.

AlbertoMonteiro:
We also need this fix!

josefpihrt:
@copilot resolve the merge conflicts in this pull request

src/CommandLine/CommandLine.csproj
```diff
@@ -59,7 +59,14 @@
...
     <PackageReference Include="Microsoft.Build" Version="17.14.28" ExcludeAssets="runtime" Condition="'$(TargetFramework)' == 'net9.0'" />
-    <PackageReference Include="Microsoft.Build" Version="18.0.2" ExcludeAssets="runtime" Condition="'$(TargetFramework)' == 'net10.0'" />
+    <PackageReference Include="Microsoft.Build" Version="18.4.0" ExcludeAssets="runtime" Condition="'$(TargetFramework)' == 'net10.0'" />
+    <!-- Suppress bundling of Microsoft.Build.Framework.dll: the NuGet copy is a
+         ref-only assembly that lacks internal types (e.g. FileUtilities) which
+         the SDK's full Microsoft.Build.dll calls when parsing .slnx solutions.
+         MSBuildLocator will load the matching full implementation from the SDK
+         at runtime. See https://github.com/dotnet/roslynator/issues/1729. -->
+    <PackageReference Include="Microsoft.Build.Framework" Version="17.14.28" ExcludeAssets="runtime" Condition="'$(TargetFramework)' == 'net9.0'" />
+    <PackageReference Include="Microsoft.Build.Framework" Version="18.4.0" ExcludeAssets="runtime" Condition="'$(TargetFramework)' == 'net10.0'" />
```
josefpihrt, line 69: The exclusion needs to cover the rest of the set, not just `Microsoft.Build.Framework`. These are still deployed, pulled in transitively by `Microsoft.CodeAnalysis.Workspaces.MSBuild`:

| Assembly | deployed version |
|---|---|
| `Microsoft.Build.Tasks.Core` | 17.7.2 |
| `Microsoft.Build.Utilities.Core` | 17.7.2 |
| `Microsoft.NET.StringTools` | 18.0.2 |

`MSBuildLocator` hooks `AssemblyLoadContext.Resolving`, which only fires when an assembly is *not* found next to the app — so any deployed copy wins and the SDK's is never consulted. Against an SDK 10 engine at 18.x, 17.7.2 copies of `Tasks.Core`/`Utilities.Core` are a whole major version apart.

Also worth noting: line 55 suppresses a known high-severity advisory for `Microsoft.Build.Tasks.Core` 17.7.2. Excluding its runtime asset stops that DLL from shipping at all — though the suppression still has to stay, since NuGetAudit scans the restore graph rather than the deployed set.

Handled in #1783 for net8.0, net9.0 and net10.0.

josefpihrt (changes requested):
Thanks for chasing this down — the shadowing diagnosis is correct, and it is what made the full fix possible. I have opened #1783, which supersedes this PR; you are credited on the commit and in the ChangeLog.

One correction, and I had the same wrong reading at first. The NuGet `lib/` assembly is not a ref-only build stripped of internals — it is a real implementation. The size gap that suggests otherwise turns out to be ReadyToRun compilation in the SDK's copy, not missing types:

```
NuGet 17.14.28 lib     361,232   R2R=no
SDK 9.0.302          1,011,200   R2R=yes    (same MSBuild version)
```

What actually happens is version drift. MSBuild moved shared helpers into `Microsoft.Build.Framework` between 18.4 and 18.6:

| `Microsoft.Build.Framework` | `lib` size | contains `FileUtilities` |
|---|---|---|
| 18.0.2 | 269,608 | no |
| 18.4.0 | 305,424 | no |
| 18.6.3 | 529,712 | **yes** |
| 18.8.2 | 1,051,472 | **yes** |

An SDK 10 patch whose engine is 18.6 or later asks for a type that the deployed 18.0.2 build does not have. Two consequences for this PR:

- The bump to 18.4.0 does not cross that boundary, so it would not have fixed the crash on its own.
- Bumping to 18.8.2 **would** have made this particular `TypeLoadException` disappear — which is the trap. MSBuild's assemblies are a matched set, so any version we deploy is wrong for someone: the user's SDK patch level is unknowable and moves independently of Roslynator's releases. Deploying none of them is the only stable answer.

It is also not `.slnx`-specific — #1729 is a `MissingFieldException` on a plain console project. That is why #1783 covers net8.0 as well: an installed tool's TFM is frozen at install time, so anyone who installed under SDK 8 or 9 and later moved to SDK 10 keeps that asset and hits the same mismatch.

#1783 excludes the runtime assets of all five MSBuild-family packages for net8.0, net9.0 and net10.0, pins versions to what the restore graph already resolves so the deployed dependency set is unchanged, and adds a build target that fails the build if any of them is ever deployed again.

One thing I could not verify: the SDK available to me is 10.0.101, which ships MSBuild 18.0.6 — before the 18.6 change. So I can confirm the assemblies are gone, but not that your `TypeLoadException` disappears. If you still have that environment, would you mind trying #1783?

Thanks again — the analysis in your description is what pointed at the right place.

josefpihrt:
fixed with #1783

PR #1772, opened 2026-06-18
Fix analyzer RCS1246 for conditional access expressions

krajek:
Smallest reproducible program I could come up with:

`new { Values = new int[0] }?.Values.ElementAt(0);`

Without the fix, the analyzer in that case throws NRE and fails the build.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix analyzer [RCS1194](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1194) ([PR](https://github.com/dotnet/roslynator/pull/1733))
 - Fix analyzer [RCS1060](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1060) to ignore classes marked with `file` modifier ([PR](https://github.com/dotnet/roslynator/pull/1777) by @cbersch)
 - Fix analyzer [RCS1231](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1231) ([PR](https://github.com/dotnet/roslynator/pull/1774) by @cbersch)
-- Fix analyzer [RCS1246](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1246) for conditional access expressions ([PR](https://github.com/dotnet/roslynator/pull/XXXX))
+- Fix analyzer [RCS1246](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1246) for conditional access expressions ([PR](https://github.com/dotnet/roslynator/pull/1772))
 - [CLI] Fix `fix` command ignoring `--include` / `--exclude` file filter ([PR](https://github.com/dotnet/roslynator/pull/1758) by @hashiiiii)
 - [CLI] Fix loading of projects and solutions on .NET 10 SDK ([PR](https://github.com/dotnet/roslynator/pull/1783))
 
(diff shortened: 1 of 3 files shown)


krajek:
@dotnet-policy-service agree

krajek:
Hi, is there anything I could do to move this forward?

ChangeLog.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix enum contained flags check for partial matches in [RCS1258](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1258) ([PR](https://github.com/dotnet/roslynator/pull/1740) by @ovska)
 - Fix analyzer [RCS1146](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1146) ([PR](https://github.com/dotnet/roslynator/pull/1747))
 - Fix analyzer [RCS1194](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1194) ([PR](https://github.com/dotnet/roslynator/pull/1733))
+- Fix analyzer [RCS1246](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1246) for conditional access expressions ([PR](https://github.com/dotnet/roslynator/pull/XXXX))
```
josefpihrt: Please fix PR number and add your name, if you wish.

src/Common/CSharp/Analysis/UseElementAccessAnalysis.cs
```diff
@@ -35,6 +35,12 @@ public static bool IsFixableElementAt(
         if (!HasAccessibleIndexer(typeSymbol, reducedExtensionMethodInfo.ReducedSymbolOrSymbol.ReturnType, semanticModel, invocationExpression.SpanStart))
             return false;
 
+        // Skip the speculative binding check for member binding expressions (conditional access context)
+        // because GetSpeculativeSymbolInfo cannot bind a MemberBindingExpression as a standalone expression.
+        // The infinite recursion scenario this check guards against is not applicable in conditional access.
+        if (invocationInfo.Expression.IsKind(SyntaxKind.MemberBindingExpression))
+            return true;
+
```
josefpihrt: The crash fix here is good, but this comment isn't quite accurate:

> The infinite recursion scenario this check guards against is not applicable in conditional access.

The scenario is reachable through conditional access — the MemberBindingExpression receiver just needs to be typed as the enclosing indexer's type. For example:

```
class C : IReadOnlyList<int>
{
    public C Nested;

    public int this[int index] => Nested?.Nested.ElementAt(index) ?? 0;
    // fix -> Nested?.Nested[index], which calls C.this[int] — the enclosing indexer
    // ...IReadOnlyList<int> members...
}
```

The non-conditional twin (Nested.Nested.ElementAt(index)) is suppressed today because the speculative bind resolves [index] to C.this and CheckInfiniteRecursion returns false. With this return true, the conditional-access form skips that check and offers the fix, which converts an enumerating ElementAt call into a direct indexer call — potential infinite recursion (a StackOverflowException for a self-referential/cyclic graph).

I think this is a fine, rare gap to accept for now rather than block the crash fix — but could we reword the comment so it doesn't claim the scenario can't happen? Something like:

// GetSpeculativeSymbolInfo cannot bind a MemberBindingExpression (conditional access,
// e.g. 'x?.Values.ElementAt(0)') as a standalone expression, so the speculative element
// access below would throw. That binding's only purpose is infinite-recursion detection,
// which we intentionally skip here. The recursion scenario is still technically reachable
// via conditional access (a member returning the enclosing indexer's type), but it's rare
// enough that we accept not guarding it in order to fix the crash.

josefpihrt (changes requested):

josefpihrt:
@krajek Hey, thank you for your contribution and sorry for late review.

src/Common/CSharp/Analysis/UseElementAccessAnalysis.cs
krajek: Got it. Thanks for the detailed review. I think I can see the tradeoff now (contrived recursion bug vs fixing a real crash). Reworded the comment.

josefpihrt (approved):

PR #1774, opened 2026-06-19
Fix analyzer RCS1231 (#1769)

### LINKED ISSUE #1769: RCS1231 shouldn't trigger when parameter is `params`
### Description
Per C# compiler message [CS1611](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/params-arrays?f1url=%3FappId%3Droslyn%26k%3Dk(CS1611)#method-declaration-rules) `ref` or `out` is not allowed on `params`. `in` isn't mentioned specifically but the compiler error triggers when declaring it on `params`.

### Example
```cs
public static int Create<T>(params ReadOnlySpan<T> values) // <- this triggers RCS1231
{
    HashCode hash = new();
    foreach (T value in values)
    {
        hash.Add(value);
    }

    return hash.ToHashCode();
}
```

cbersch:
Fix #1769

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix analyzer [RCS1231](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1231) ([PR](https://github.com/dotnet/roslynator/pull/1744) by @cbersch)
+- Fix analyzer [RCS1231](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1231) ([PR](https://github.com/dotnet/roslynator/pull/1774) by @cbersch)
 - Fix enum contained flags check for partial matches in [RCS1258](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1258) ([PR](https://github.com/dotnet/roslynator/pull/1740) by @ovska)
 - Fix analyzer [RCS1146](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1146) ([PR](https://github.com/dotnet/roslynator/pull/1747))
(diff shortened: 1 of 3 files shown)


ChangeLog.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix analyzer [RCS1231](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1231) ([PR](https://github.com/dotnet/roslynator/pull/1744) by @cbersch)
```
josefpihrt:
```suggestion
- Fix analyzer [RCS1231](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1231) ([PR](https://github.com/dotnet/roslynator/pull/1774) by @cbersch)
```

src/Tests/Analyzers.Tests/RCS1231MakeParameterRefReadOnlyTests.cs
```diff
@@ -122,6 +122,36 @@ void M(C c, C c)
...
 
+    [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.MakeParameterRefReadOnly)]
+    public async Task TestNoDiagnostic_ParamsArray()
+    {
+        await VerifyNoDiagnosticAsync(@"
+using System;
+
+readonly struct C
+{
+    void M(params C[] c)
```
josefpihrt: Array is not a struct type so this test doesn't exercise the fix.

josefpihrt (changes requested):

src/Tests/Analyzers.Tests/RCS1231MakeParameterRefReadOnlyTests.cs
cbersch: Right, this test is useless. I'll remove it.

josefpihrt (approved):

PR #1777, opened 2026-07-08
Fix analyzer RCS1060

### LINKED ISSUE #1775: RCS1060 fires even when I have an explicit `file` scope modifier on a class.
If I've specifically created a file-specific class using the `file` modifier, I'd expect RCS1060 not to fire as I've made an explicit design choice.

cbersch:
* Ignore classes with `file` modifier, because that expresses an explicit design choice which must be equivalent to ignoring the diagnostic.
Fix #1775

### DIFF
--- a/src/Analyzers/CSharp/Analysis/DeclareEachTypeInSeparateFileAnalyzer.cs
+++ b/src/Analyzers/CSharp/Analysis/DeclareEachTypeInSeparateFileAnalyzer.cs
@@ -60,7 +60,11 @@ public sealed class DeclareEachTypeInSeparateFileAnalyzer : BaseDiagnosticAnalyz
                 {
                     Analyze(namespaceDeclaration.Members);
                 }
-                else if (SyntaxFacts.IsTypeDeclaration(member.Kind()) && !member.Modifiers.Contains(SyntaxKind.FileKeyword))
+                else if (SyntaxFacts.IsTypeDeclaration(member.Kind())
+#if ROSLYN_4_4
+                    && !member.Modifiers.Contains(SyntaxKind.FileKeyword)
+#endif
+                    )
                 {
                     if (firstTypeDeclaration is null)
                     {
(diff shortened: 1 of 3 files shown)


src/Tests/Analyzers.Tests/RCS1060DeclareEachTypeInSeparateFileTests.cs
```diff
@@ -16,45 +16,69 @@ public class RCS1060DeclareEachTypeInSeparateFileTests : AbstractCSharpDiagnosti
     public async Task Test_Namespace()
     {
         await VerifyDiagnosticAndFixAsync("""
-namespace N
-{
-    public class [|C1|]
-    {
-    }
+            namespace N
```
josefpihrt: Please do not indent test code, it's established format in the tests and it also improves readability for test code with longer lines.

src/Analyzers/CSharp/Analysis/DeclareEachTypeInSeparateFileAnalyzer.cs
```diff
@@ -60,7 +60,7 @@ void Analyze(SyntaxList<MemberDeclarationSyntax> members)
                 {
                     Analyze(namespaceDeclaration.Members);
                 }
-                else if (SyntaxFacts.IsTypeDeclaration(member.Kind()))
+                else if (SyntaxFacts.IsTypeDeclaration(member.Kind()) && !member.Modifiers.Contains(SyntaxKind.FileKeyword))
```
josefpihrt: Code build against multiple versions of Roslyn and `file` modifier is not available in some versions, please add conditional compilation:

```
#if ROSLYN_4_4
    && !member.Modifiers.Contains(SyntaxKind.FileKeyword)
#endif
```

josefpihrt (changes requested):

josefpihrt:
@cbersch fix looks fine, I just added a couple of comments.

src/Tests/Analyzers.Tests/RCS1060DeclareEachTypeInSeparateFileTests.cs
cbersch: Sure, sorry.

josefpihrt (approved):

josefpihrt:
@cbersch Thank you for your contribution!

PR #1782, opened 2026-08-01
Fix RCS1118 to not report local variable passed as 'in' argument

### LINKED ISSUE #1781: RCS1118 suggests `const` for a local variable that is passed as an `in` argument
**Product and Version Used**: Roslynator.Analyzers 4.15.0 (also reproduces on current `main`)

**Steps to Reproduce**:

RCS1118 reports a local variable that is later passed to a parameter with an explicit `in` argument keyword. Applying the fix produces code that does not compile, because a constant cannot be passed by readonly reference.

```csharp
class C
{
    void M()
    {
        int repeat = 0; // RCS1118 reported here
        Bar(in repeat);
    }

    void Bar(in int p)
    {
    }
}
```

The same happens when the parameter is `ref readonly` instead of `in`.

I originally hit this with Silk.NET:

```csharp
int repeat = (int)GLEnum.Repeat;
_gl.TexParameterI(TextureTarget.Texture2D, TextureParameterName.TextureWrapS, in repeat);
```

**Actual Behavior**:

RCS1118 is reported, and the code fix produces code that fails to compile with `error CS8156: An expression cannot be used in this context because it may not be passed or returned by reference`:

```csharp
class C
{
    void M()
    {
        const int repeat = 0;
        Bar(in repeat); // error CS8156
    }

    void Bar(in int p)
    {
    }
}
```

**Expected Behavior**:

No diagnostic is reported, since the local variable cannot be marked as const:

```csharp
class C
{
    void M()
    {
        int repeat = 0;
        Bar(in repeat);
    }

    void Bar(in int p)
    {
    }
}
```

---

**Cause** (in case it is useful):

`MarkLocalVariableAsConstWalker` inherits `VisitArgument` from `AssignedExpressionWalker`, which only handles `ref` and `out`, so an `in` argument is not seen as a usage that prevents `const`. `UseConstantInsteadOfFieldWalker` (RCS1187) already has an explicit `SyntaxKind.InKeyword` case for the same situation.

I'd be happy to submit a PR for this. *EDIT: See https://github.com/dotnet/roslynator/pull/1782*

**Note on a related case that is _not_ part of this report**: when a `ref readonly` parameter is called without an argument keyword (`Bar(repeat)`), marking the local as const does not break the build; it only changes warning CS9192 to CS9193. Only an explicit `in`/`ref` at the call site is a hard error.

NoahStolk:
Fixes #1781

RCS1118 reports a local variable that is later passed to a parameter with an explicit `in` argument keyword. Marking such a local as const produces code that does not compile, because a constant cannot be passed by readonly reference.

Added two new tests: one for an `in` parameter and one for a `ref readonly` parameter. Both fail on `main` and pass with the fix.

### DIFF
--- a/src/Tests/Analyzers.Tests/RCS1118MarkLocalVariableAsConstTests.cs
+++ b/src/Tests/Analyzers.Tests/RCS1118MarkLocalVariableAsConstTests.cs
@@ -158,4 +158,34 @@ public static class C
     }
 }");
     }
+
+    [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.MarkLocalVariableAsConst)]
+    public async Task Test_InParameter_WithoutInKeyword()
+    {
+        await VerifyDiagnosticAndFixAsync(@"
+public static class C
+{
+    static void Foo()
+    {
+        [|int|] x = 0;
+        Bar(x);
+    }
+
+    public static void Bar(in int p)
+    {
+    }
+}", @"
+public static class C
+{
+    static void Foo()
+    {
+        const int x = 0;
+        Bar(x);
+    }
+
+    public static void Bar(in int p)
+    {
+    }
+}");
+    }
 }
--- a/src/Tests/Analyzers.Tests/RCS1118MarkLocalVariableAsConstTests.cs
+++ b/src/Tests/Analyzers.Tests/RCS1118MarkLocalVariableAsConstTests.cs
@@ -122,4 +122,40 @@ public static class C
     }
 }");
     }
+
+    [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.MarkLocalVariableAsConst)]
+    public async Task TestNoDiagnostic_InArgument()
+    {
+        await VerifyNoDiagnosticAsync(@"
+public static class C
+{
+    static void Foo()
+    {
+        int x = 0;
+        Bar(in x);
+    }
+
+    public static void Bar(in int p)
+    {
+    }
+}");
+    }
+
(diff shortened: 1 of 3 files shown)


src/Tests/Analyzers.Tests/RCS1118MarkLocalVariableAsConstTests.cs
```diff
@@ -120,6 +120,42 @@ public static int Bar(this ref int p1, int p2)
     {
         return p1 + p2;
     }
+}");
+    }
+
+    [Fact, Trait(Traits.Analyzer, DiagnosticIdentifiers.MarkLocalVariableAsConst)]
+    public async Task TestNoDiagnostic_InArgument()
```
josefpihrt, line 127: Suggestion: consider adding a *positive* test asserting the diagnostic still reports when a const-eligible local is passed to an `in` parameter **without** the `in` keyword, e.g.:

```csharp
int x = 0;   // still reported: [|int x = 0;|]
Bar(x);      // Bar(in int p)
```

That compiles today (the compiler synthesizes a temporary), so RCS1118 should still fire — and the test would lock in that this fix is scoped to the explicit keyword, guarding against a future over-suppression regression.

josefpihrt (changes requested):
Please update change log.

src/Tests/Analyzers.Tests/RCS1118MarkLocalVariableAsConstTests.cs
NoahStolk: Good suggestion, I've added `Test_InParameter_WithoutInKeyword`.

josefpihrt (approved):

josefpihrt:
@NoahStolk Thank you for your contribution!

PR #1787, opened 2026-08-08
Bump Roslyn to 5.0

josefpihrt:
## Summary

Breaking release: migrate Roslynator to **Roslyn 5.0.0** and retarget IDE extensions for the Roslyn 5 toolchain.

- Bump default build, CLI, and testing to Roslyn **5.0.0**
- Add **`roslyn5.0`** NuGet package flavor (`analyzers/dotnet/roslyn5.0/cs`) alongside existing `roslyn3.8` and `roslyn4.7`
- Replace **Roslynator 2022** VSIX with **Roslynator 2026** for Visual Studio 2026 (`[18.0,19.0)`)
- Remove bundled analyzers and RCS analyzer code fixes from Visual Studio and VS Code extensions (use `Roslynator.Analyzers` NuGet for diagnostics and RCS fixes)
- IDE extensions ship **refactorings** and **compiler diagnostic code fixes** only

## Breaking changes

| Area | Before | After |
|------|--------|-------|
| Roslyn toolchain | 4.14.0 (default) | **5.0.0** |
| VS extension | Roslynator 2022, bundled analyzers | **Roslynator 2026**, refactorings + CS fixes only |
| VS Code extension | Bundled analyzers + refactorings + fixes | Refactorings + CS fixes; analyzers/RCS fixes via NuGet |
| VS 2022 VSIX | Supported | **Not supported** — pin 4.x or use NuGet |

## Extension behavior

**Visual Studio 2026**
- Analyzers and RCS analyzer code fixes: `Roslynator.Analyzers` NuGet
- Refactorings and CS code fixes: extension

**VS Code** (OmniSharp required; `dotnet.server.useOmnisharp`: `true`)
- Analyzers and RCS analyzer code fixes: NuGet
- Refactorings + CS code fixes: extension
- Requires C# extension with OmniSharp **1.39.15+** (Roslyn 5.x)

## Technical notes

- Roslyn 5 API fixes: `GetDeclaredLocalSymbol`, `RegisterWorkspaceFailedHandler`, null-safety updates
- CLI net8.0: `Microsoft.Build` 17.11.31 minimum for Roslyn 5 `Workspaces.MSBuild`
- VSSDK: `Microsoft.VisualStudio.SDK` 17.14 + `Microsoft.VSSDK.BuildTools` 18.9
- `RS1038` suppressed on analyzer projects (VB-capable assemblies; tech debt)

## Test plan

- [x] `dotnet test` on Roslyn 5.0
- [x] CI: multi-flavor NuGet pack (`roslyn3.8`, `roslyn4.7`, `roslyn5.0`)
- [x] CI: `build_vs_extension` (VS 2026 VSIX)
- [ ] CI: full workflow green on latest push
- [ ] Manual: VS 2026 extension — refactorings light-bulb
- [ ] Manual: VS Code + OmniSharp — refactorings light-bulb
- [ ] Publish Roslynator 2026 marketplace listing after merge


src/Tools/CodeGeneration/Markdown/MarkdownGenerator.cs
```diff
@@ -256,7 +256,7 @@ static IEnumerable<MElement> CreateAppliesTo(AnalyzerMetadata analyzer)
 
             if (!analyzer.Id.StartsWith("RCS9"))
             {
-                yield return BulletItem(Link("Extension for VS 2022", "https://marketplace.visualstudio.com/items?itemName=josefpihrt.Roslynator2022"));
+                yield return BulletItem(Link("Extension for VS 2026", "https://marketplace.visualstudio.com/items?itemName=josefpihrt.Roslynator2026"));
                 yield return BulletItem(Link("Extension for VS Code", "https://marketplace.visualstudio.com/items?itemName=josefpihrt-vscode.roslynator"));
                 yield return BulletItem(Link("Extension for Open VSX", "https://open-vsx.org/extension/josefpihrt-vscode/roslynator"));
```
josefpihrt: **Analyzer docs still claim IDE-extension support after this PR removes analyzers from both extensions.**

`CreateAppliesTo` still emits VS 2026 / VS Code / Open VSX bullets for every non-RCS9 analyzer. After this change those extensions no longer ship analyzers (NuGet only). Generated analyzer pages would tell users to install the VSIX/VS Code extension to get RCS0/RCS1 diagnostics, which will not work.

Drop the three extension bullets here (keep the NuGet package links), or gate them on a product that actually still bundles analyzers.

src/Analyzers.CodeFixes/docs/NuGetReadme.md
```diff
@@ -4,7 +4,7 @@ A collection of 200+ [analyzers](https://josefpihrt.github.io/docs/roslynator/an
 
 ## Requirements
 
-* Visual Studio 2022
+* Visual Studio 2026
```
josefpihrt: **NuGet packages still support VS 2022 / Roslyn 3.8+; this requirement line should not become “Visual Studio 2026”.**

The package still ships `roslyn3.8`, `roslyn4.7`, and `roslyn5.0` folders. VS 2022 (and `dotnet build` on older SDKs) should keep working via the existing flavors. Replacing the requirement with VS 2026 implies the NuGet analyzers no longer work on VS 2022, which is the opposite of the migration story (VSIX dropped; NuGet is the VS 2022 path).

Same edit is copied across the other `docs/NuGetReadme.md` files. Prefer something like “Visual Studio 2022 or 2026” (or leave VS 2022) and keep “Roslyn 3.8.0 or higher” as the real floor.

src/VisualStudio/source.extension.vsixmanifest
```diff
@@ -36,23 +36,13 @@
...
-    <Asset Type="Microsoft.VisualStudio.Analyzer" d:Source="Project" d:ProjectName="CSharp.Workspaces" Path="|CSharp.Workspaces|" />
     <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="CSharp.Workspaces" Path="|CSharp.Workspaces|" />
-    <Asset Type="Microsoft.VisualStudio.Analyzer" d:Source="Project" d:ProjectName="Workspaces.Common" Path="|Workspaces.Common|" />
     <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="Workspaces.Common" Path="|Workspaces.Common|" />
-    <Asset Type="Microsoft.VisualStudio.Analyzer" d:Source="Project" d:ProjectName="Workspaces.Core" Path="|Workspaces.Core|" />
     <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="Workspaces.Core" Path="|Workspaces.Core|" />
-    <Asset Type="Microsoft.VisualStudio.Analyzer" d:Source="Project" d:ProjectName="Analyzers" Path="|Analyzers|" />
-    <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="Analyzers" Path="|Analyzers|" />
     <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="Analyzers.CodeFixes" Path="|Analyzers.CodeFixes|" />
```
josefpihrt: **VSIX still MEF-exports `Analyzers.CodeFixes` / `Formatting.Analyzers.CodeFixes` after removing the analyzer DLLs.**

Recommended setup is now: diagnostics from `Roslynator.Analyzers` (which already includes those code-fix assemblies) plus this VSIX. That registers the same RCS/formatting fix providers twice in VS 2026 (duplicate light-bulb items / MEF noise). VS Code already dropped both the analyzer and analyzer-code-fix DLLs, which is the consistent model.

If RCS fixes in the IDE are supposed to come only from NuGet, remove these two MefComponent assets and the matching `ProjectReference`s. If the VSIX is intentionally the only source of RCS fixes, say that in Overview.md and do not also pack code fixes in the analyzer NuGet for the VS 2026 flavor.

src/Workspaces.Core/CodeFixes/DiagnosticFixProvider.cs
```diff
@@ -202,6 +202,7 @@ public static async Task<DiagnosticFix> GetFixAsync(
             (a, _) =>
             {
                 if (!equivalenceKeys.IsDefaultOrEmpty
+                    && a.EquivalenceKey is not null
                     && !equivalenceKeys.Contains(a.EquivalenceKey, StringComparer.Ordinal))
```
josefpihrt: **Null `EquivalenceKey` now bypasses the CLI equivalence-key filter.**

`StringComparer.Ordinal` throws on null, so the new `a.EquivalenceKey is not null` guard is necessary — but combined with `&&`, a null-key action is no longer rejected when `equivalenceKeys` is set. `roslynator fix` with a specific equivalence key can therefore pick a different/unkeyed action.

Treat null as non-matching:

```csharp
if (!equivalenceKeys.IsDefaultOrEmpty
    && (a.EquivalenceKey is null
        || !equivalenceKeys.Contains(a.EquivalenceKey, StringComparer.Ordinal)))
{
    return;
}
```

josefpihrt (commented):
Detailed review of the Roslyn 5.0 / Roslynator 2026 migration. The toolchain bump, `roslyn5.0` pack flavor, `GetDeclaredLocalSymbol` helper, `RegisterWorkspaceFailedHandler`, and dropping bundled analyzers from both IDEs are directionally right. A few product/docs mismatches should be fixed before this leaves draft — especially analyzer “Applies to” docs and NuGet requirements still describing the old VSIX-includes-analyzers world.

Also: this branch last moved on 8 Aug and will need a rebase onto current `main` (extension-block / `SyntaxKind.ExtensionDeclaration` work in particular).

src/CSharp/CSharp/Extensions/CSharpExtensions.cs
```diff
@@ -25,6 +25,17 @@ public static class CSharpExtensions
...
 
+    internal static ILocalSymbol? GetDeclaredLocalSymbol(
+        this SemanticModel semanticModel,
+        VariableDesignationSyntax designation,
+        CancellationToken cancellationToken = default)
+    {
+        if (designation is SingleVariableDesignationSyntax singleVariableDesignation)
+            return (ILocalSymbol?)semanticModel.GetDeclaredSymbol(singleVariableDesignation, cancellationToken);
+
+        return null;
```
josefpihrt, line 36: **`GetDeclaredLocalSymbol` returns null for `ParenthesizedVariableDesignationSyntax` (`var (x, y)`).**

That is enough to compile against Roslyn 5 (no `GetDeclaredSymbol(VariableDesignationSyntax)`), but `ChangeDeclarationExpressionTypeRefactoring` now skips `UseExplicitType` whenever the designation is not a single variable — even though `CSharpTypeAnalysis.IsImplicitThatCanBeExplicit` still returns true for deconstruction (it walks child `SingleVariableDesignation`s).

The analyzer code-fix path already falls back to `GetTypeSymbol(declarationExpression)`. The refactoring should do the same (or this helper should resolve through parenthesized/discards) so `var (a, b) = ...` does not lose the explicit-type refactoring.

src/Directory.Build.props
```diff
@@ -20,19 +20,33 @@
...
+        <RoslynatorRoslynVersion>5.0.0</RoslynatorRoslynVersion>
+        <RoslynatorRoslynAnalyzersVersion>3.11.0</RoslynatorRoslynAnalyzersVersion>
+        <DefineConstants>$(DefineConstants);ROSLYN_3_8;ROSLYN_4_0;ROSLYN_4_2;ROSLYN_4_4;ROSLYN_4_7;ROSLYN_5_0;ROSLYN_5_0_ONLY</DefineConstants>
+      </PropertyGroup>
+    </When>
+    <Otherwise>
+      <PropertyGroup>
+        <RoslynatorRoslynVersion>5.0.0</RoslynatorRoslynVersion>
+        <RoslynatorRoslynAnalyzersVersion>3.11.0</RoslynatorRoslynAnalyzersVersion>
+        <DefineConstants>$(DefineConstants);ROSLYN_3_8;ROSLYN_4_0;ROSLYN_4_2;ROSLYN_4_4;ROSLYN_4_7;ROSLYN_5_0</DefineConstants>
```
josefpihrt: **Default (Otherwise) Roslyn 5.0 build does not define `ROSLYN_5_0_ONLY`, while `RoslynVersion=roslyn5.0` does.**

Previously the default Otherwise matched the then-current flavor including `ROSLYN_4_7_ONLY`. Nothing in-tree uses `ROSLYN_*_ONLY` today, so this is latent — but any future `#if ROSLYN_5_0_ONLY` will silently not apply to the solution default build. Consider making Otherwise identical to the `roslyn5.0` `When` block (or drop `_ONLY` from the named flavor too).

src/Tools/CodeGeneration/Markdown/MarkdownGenerator.cs
josefpihrt: Addressed: dropped the three IDE extension bullets from `CreateAppliesTo`; analyzer pages now list NuGet packages only.

src/Analyzers.CodeFixes/docs/NuGetReadme.md
josefpihrt: Addressed: NuGet requirements now say “Visual Studio 2022 or 2026” (Roslyn 3.8+ floor unchanged). Related Products still link to Roslynator 2026.

src/VisualStudio/source.extension.vsixmanifest
josefpihrt: Addressed: removed `Analyzers.CodeFixes` / `Formatting.Analyzers.CodeFixes` MefComponents and project refs from the VSIX. RCS diagnostics/fixes come from NuGet only (same model as VS Code). Extension keeps refactorings + CS `CodeFixes`.

src/Workspaces.Core/CodeFixes/DiagnosticFixProvider.cs
josefpihrt: Addressed: null `EquivalenceKey` is now treated as non-matching when an equivalence-key filter is set.

src/CSharp/CSharp/Extensions/CSharpExtensions.cs
josefpihrt: Addressed: `ChangeDeclarationExpressionTypeRefactoring` falls back to `GetTypeSymbol(declarationExpression)` when the designation is not a single variable (same as the analyzer code-fix path).

src/Directory.Build.props
josefpihrt: Addressed: `Otherwise` now defines `ROSLYN_5_0_ONLY` like the `roslyn5.0` flavor.

PR #1791, opened 2026-08-08
fix(analyzers): fix RCS0034 for primary constructor types

### LINKED ISSUE #1761: RCS0034 incorrectly merges first 'where' clause onto declaration line for generic type with primary constructor and multiple type parameters
### Steps to Reproduce

Given the following code:

```csharp
internal sealed class Example(TEntry entry)
    where TEntry : class
    where TValue : class
{
    public required TEntry Entry { get; set; } = entry;
}
```

RCS0034 (`PutTypeParameterConstraintOnItsOwnLine`) is reported on the first `where` clause, and applying its code fix produces:

```csharp
internal sealed class Example(TEntry entry) where TEntry : class
    where TValue : class
{
    public required TEntry Entry { get; set; } = entry;
}
```

### Expected Behavior

No diagnostic should be reported. Each `where` clause is already on its own line, which is exactly what RCS0034 is supposed to enforce. The result of the fix is also internally inconsistent: it puts the first constraint inline while leaving the second on its own line.

### Actual Behavior

RCS0034 fires on the first `where` clause and the code fix moves it to the declaration line, producing inconsistent formatting.

### Conditions Required to Reproduce

The bug appears specifically when **all** of the following hold:

1. The type has a **primary constructor** (`class Foo<...>(...)`).
2. There are **two or more type parameters**.
3. There are **two or more `where` constraint clauses**, each already on its own line.
4. The type has **no base type or interface list**.

If any one of these is changed (single type parameter, single constraint, or a base/interface list is present), RCS0034 behaves correctly.

### Related Issue

This looks like a regression / incomplete fix of #1347, which addressed the same analyzer for the case with an interface in the base list. The no-base-list + multiple-constraints case appears to have been missed.

### Workaround

Suppressing the rule per-file or globally via `.editorconfig`:
dotnet_diagnostic.RCS0034.severity = none

### Additional Notes

- I have reproduced this on the latest released version of Roslynator analyzers.
- Happy to provide a failing analyzer test case or open a PR if it helps.

josefpihrt:
## Summary

- RCS0034 no longer reports a false positive when a type has a primary constructor and multiple `where` clauses on separate lines
- Anchor constraint trivia on the primary constructor closing parenthesis instead of the type parameter list

Fixes #1761

## Test plan

- [x] `dotnet test src/Tests/Formatting.Analyzers.Tests/Formatting.Analyzers.Tests.csproj --filter FullyQualifiedName~RCS0034PutTypeParameterConstraintOnItsOwnLineTests`

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix analyzer [RCS0034](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0034) for types with a primary constructor and multiple constraint clauses
+
 ## [4.16.0] - 2026-08-08
 
 ### Added
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix analyzer [RCS0034](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0034) for types with a primary constructor and multiple constraint clauses
+- Fix analyzer [RCS0034](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0034) for types with a primary constructor and multiple constraint clauses ([PR](https://github.com/dotnet/roslynator/pull/1791))
 
 ## [4.16.0] - 2026-08-08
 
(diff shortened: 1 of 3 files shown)


ChangeLog.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix analyzer [RCS0034](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0034) for types with a primary constructor and multiple constraint clauses
```
josefpihrt: Add PR link to the changelog entry.

```suggestion
- Fix analyzer [RCS0034](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS0034) for types with a primary constructor and multiple constraint clauses ([PR](https://github.com/dotnet/roslynator/pull/1791))
```

PR #1793, opened 2026-08-09
docs: polish hand-written documentation prose

josefpihrt:
## Summary
- Polish grammar and rephrasing across README, VS Code README, Visual Studio overviews, NuGet readmes, CLI generator fragments, and CONTRIBUTING.md
- Add inline code formatting for technical identifiers in prose (flags, filenames, diagnostic ID patterns, EditorConfig)
- Update outdated GitHub URLs (`josefpihrt/orang`, `josefpihrt/Wordb`, sponsors username); changelogs and `JosefPihrt/Roslynator` config paths are unchanged

## Test plan
- [ ] Review README and VS Code marketplace README for accuracy
- [ ] Spot-check NuGet readme rendering on nuget.org after next package publish
- [ ] Verify CLI documentation generator output if regenerated from `data/*.md` fragments

### DIFF
--- a/src/CommandLine.DocumentationGenerator/data/rename-symbol_bottom.md
+++ b/src/CommandLine.DocumentationGenerator/data/rename-symbol_bottom.md
@@ -1,6 +1,6 @@
 ## Redirected/Piped Input
 
-Redirected/piped input will be used as a list of project/solution paths separated with newlines.
+Redirected or piped input is treated as a list of project/solution paths, one per line.
 
 ## Examples
 
@@ -103,7 +103,7 @@ roslynator rename-symbol my.sln
 #### Notes
 
 * We assume that there are files `my.sln` and `rename-constants.cs` in the current directory.
-* If the "IsMatch" method and "GetNewName" method are defined both in the same file it is allowed to specified it only once as a value of `--match-from` option.
+* If the `IsMatch` method and `GetNewName` method are defined both in the same file it is allowed to specified it only once as a value of the `--match-from` option.
 * Because the code is marked with `auto-generated` tag we have to explicitly allow generated code with `--include-generated-code` switch.
 * The option `--scope member` is not required but it can make the process faster.
 
--- a/src/CommandLine.DocumentationGenerator/data/rename-symbol_bottom.md
+++ b/src/CommandLine.DocumentationGenerator/data/rename-symbol_bottom.md
@@ -103,7 +103,7 @@ roslynator rename-symbol my.sln
 #### Notes
 
 * We assume that there are files `my.sln` and `rename-constants.cs` in the current directory.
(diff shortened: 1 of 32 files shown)


src/CommandLine.DocumentationGenerator/data/spellcheck_bottom.md
```diff
@@ -1,28 +1,28 @@
...
 
 ## Output
 
-* Command output contains up to four lists in a following order:
-  * Words containing unknown words - for example a method name that comprises multiple words where one or more of them is unknown such as `GetMaxWidht`.
-  * Unknown words - List of words that were not found in any wordlist.
-  * Auto fixes - List of automatically applied fixes.
-  * User-applied fixes - List of fixes applied by the user (when `--interactive` is set).
+* Command output contains up to four lists in the following order:
+  * Lines containing unknown words — for example, a method name made up of multiple words where one or more is unknown, such as `GetMaxWidht`.
```
josefpihrt: `Words` -> `Lines` changes the meaning here and no longer matches the CLI output.

`SpellcheckCommand` prints the literal header `Words containing unknown words:`:

https://github.com/dotnet/roslynator/blob/main/src/CommandLine/Commands/SpellcheckCommand.cs#L147

The other three bullets in this list (`Unknown words`, `Auto fixes`, `User-applied fixes`) all still match their `WriteLine` headers verbatim, so this one is now the odd one out. The example given (`GetMaxWidht`) is also a single identifier composed of several words, not a line.

Suggested revert of just the first token:

```suggestion
  * Words containing unknown words — for example, a method name made up of multiple words where one or more is unknown, such as `GetMaxWidht`.
```

src/CommandLine.DocumentationGenerator/data/rename-symbol_bottom.md
```diff
@@ -103,7 +103,7 @@ roslynator rename-symbol my.sln
 #### Notes
 
 * We assume that there are files `my.sln` and `rename-constants.cs` in the current directory.
-* If the "IsMatch" method and "GetNewName" method are defined both in the same file it is allowed to specified it only once as a value of `--match-from` option.
+* If the `IsMatch` method and `GetNewName` method are defined both in the same file it is allowed to specified it only once as a value of the `--match-from` option.
```
josefpihrt: This line got the inline-code and article treatment but the grammar was left broken: "are defined both in the same file it is allowed to specified it only once".

The underlying behaviour is correct (`Program.cs` falls back to `MatchFrom` when neither `--new-name` nor `--new-name-from` is given), so only the wording needs fixing:

```suggestion
* If the `IsMatch` and `GetNewName` methods are both defined in the same file, the file can be specified just once, as the value of the `--match-from` option.
```

Unrelated pre-existing bug in the same example, if you want to pick it up while you are here: the command above uses `--match "rename-constants.cs"`, but `--match` expects a C# expression, not a file path. Given the note about `--match-from` right below it, that should be `--match-from "rename-constants.cs"`.

PR #1794, opened 2026-08-09
chore: migrate issue template to GitHub form templates

josefpihrt:
## Summary
- Replace legacy `.github/ISSUE_TEMPLATE.md` with GitHub form templates under `.github/ISSUE_TEMPLATE/`
- Add bug report template with fields for product/version, rule ID, repro steps, actual/expected behavior, and stack traces
- Add feature request template with summary, optional rule ID, description, and prior-discussion link
- Add `config.yml` with link to CONTRIBUTING and blank issues disabled

## Test plan
- [ ] Open "New issue" on GitHub and confirm Bug report and Feature request templates appear
- [ ] Verify template fields match [CONTRIBUTING.md](CONTRIBUTING.md) guidance


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,39 @@
+name: Feature request
+description: Suggest a new feature, analyzer, refactoring, or code fix
+title: "[Feature]: "
+labels:
+  - enhancement
```
josefpihrt: This applies the wrong label for this repo. `Enhancement` exists but is effectively dead — 2 issues ever — while `Feature Request` is the label actually used, with 213 issues.

```suggestion
  - Feature Request
```

(Label names are also capitalised in this repo, so matching the exact casing avoids relying on GitHub's case-insensitive lookup.)

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,58 @@
+name: Bug report
+description: Report incorrect analyzer, refactoring, code fix, or IDE behavior
+title: "[Bug]: "
+labels:
+  - bug
```
josefpihrt: The repo's label is `Bug` (523 issues), capitalised. GitHub resolves label names case-insensitively so this will very likely still attach, but matching the existing casing costs nothing and removes the doubt.

```suggestion
  - Bug
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,58 @@
+name: Bug report
+description: Report incorrect analyzer, refactoring, code fix, or IDE behavior
+title: "[Bug]: "
```
josefpihrt: The `[Bug]: ` prefix works against this repo's established title convention. Every recent issue leads with the rule ID:

> `RCS1118 suggests \`const\` for a local variable that is passed as an \`in\` argument`
> `RCS1052 code fix drops the 'property:' attribute target on positional records`
> `RCS1060 fires for partial classes in the same file`

CONTRIBUTING.md also asks for exactly that ("**DO** include analyzer/refactoring/error ID in a title"), and the markdown blurb below repeats the request. With this prefix the ID gets pushed past a constant 7-character prefix on every issue, which makes the issue list and search-by-prefix noticeably worse for a repo where the ID is the primary sort key.

The `Bug` label already communicates the category, so I'd drop the prefix entirely:

```suggestion
```

Same reasoning applies to `title: "[Feature]: "` in `feature_request.yml`.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,58 @@
...
+    validations:
+      required: true
+  - type: textarea
+    id: expected
+    attributes:
+      label: Expected behavior
+      description: What did you expect to happen?
+      render: csharp
+    validations:
+      required: true
```
josefpihrt, line 50: `render: csharp` on these two fields fights their own descriptions. `render` wraps the *entire* textarea in a code block, so "What happened?" and "What did you expect to happen?" get syntax-highlighted as C#, and the reporter has no way to write a sentence of explanation alongside the code sample. Per the docs, `render` also disables file attachments and Markdown editing for the field — that removes the most natural place to drop a screenshot, which matters for the IDE-extension bugs this template targets.

The old `ISSUE_TEMPLATE.md` avoided this by keeping prose and a fenced ```csharp block in the same section.

Two options: drop `render` from both and let people use fences (the description already says "Include example code if possible"), or keep `render: csharp` and retitle the fields to "Actual behavior (code)" / "Expected behavior (code)" plus a separate free-form field. I'd go with the first, since it is the smaller change:

```suggestion
  - type: textarea
    id: actual
    attributes:
      label: Actual behavior
      description: What happened? Include example code if possible.
    validations:
      required: true
  - type: textarea
    id: expected
    attributes:
      label: Expected behavior
      description: What did you expect to happen?
    validations:
      required: true
```

`render: shell` on the stack-trace field below is a good fit and should stay.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
```
josefpihrt: Disabling blank issues leaves questions with nowhere to go. The repo has 18 issues labelled `Question` and Discussions is enabled, but neither template fits a question and the only contact link points at CONTRIBUTING.md, which does not mention Discussions either.

The likely outcome is that people file questions as "Bug report" — and that template makes "Steps to reproduce", "Actual behavior", and "Expected behavior" all mandatory, so they will fill them with filler text to get past validation.

Adding a Discussions link gives them the right destination:

```suggestion
blank_issues_enabled: false
contact_links:
  - name: Question or general discussion
    url: https://github.com/dotnet/roslynator/discussions
    about: Ask a question or discuss an idea before opening an issue.
  - name: Contributing guide
    url: https://github.com/dotnet/roslynator/blob/main/CONTRIBUTING.md
    about: Read issue and PR guidelines before opening a ticket.
```

josefpihrt: i'm fine with allowing blank issue, but i think it also make sense to include link to. discussions to the template. it that possible

.github/ISSUE_TEMPLATE/bug_report.yml
josefpihrt: I'd go with the first option

.github/ISSUE_TEMPLATE/bug_report.yml
josefpihrt: I agree that rule ID must be present for a bug report. But also it seems reasonable to include "Bug" or "Feature" in the title.

PR #1797, opened 2026-08-09
fix(analyzers): skip AsyncSuffix analysis when severity is suppressed

### LINKED ISSUE #1751: AsyncSuffixAnalyzer takes much time even though it's not active
Looking at the analyzer summary of our build I noticed, that the AsyncSuffixAnalyzer takes much time even though it's not active.

Image

The `.editorconfig` contains `dotnet_analyzer_diagnostic.category-roslynator.severity = none` and no explicit diagnostic for any of RCS1046, RCS1047 or RCS1047FadeOut.

The analyzer contains an `IsEffective` `if` clause, but others do as well, but don't have such an impact on the performance.

BTW: Couldn't the `Initialize` check the `startContext.Compilation` like
```C#
public override void Initialize(AnalysisContext context)
{
    base.Initialize(context);

    context.RegisterCompilationStartAction(startContext =>
    {
        if (DiagnosticHelpers.IsAnyEffective(
            startContext.Compilation,
            DiagnosticRules.AsynchronousMethodNameShouldEndWithAsync,
            DiagnosticRules.NonAsynchronousMethodNameShouldNotEndWithAsync))
        {
            startContext.RegisterSyntaxNodeAction(AnalyzeMethodDeclaration, SyntaxKind.MethodDeclaration);
        }
    });
}
```
or wouldn't that work (I don't know how `#pragma warning enable RCS1046` actually is taken into account)

josefpihrt:
## Summary
- Use syntax-tree-aware `IsEffective` checks in `AsyncSuffixAnalyzer` so RCS1046/RCS1047 skip semantic analysis when their diagnostics are suppressed via EditorConfig.
- Honor category-level severity (`dotnet_analyzer_diagnostic.category-roslynator.severity`) when determining whether analysis should run.
- Add RCS1047 tests for per-rule and category `none` severity settings.

Fixes #1751

## Test plan
- [x] `dotnet test src/Tests/Analyzers.Tests --filter FullyQualifiedName~RCS1047`
- [x] `dotnet test src/Tests/Analyzers.Tests --filter FullyQualifiedName~RCS1046`

### DIFF
--- a/src/Core/Extensions/DiagnosticsExtensions.cs
+++ b/src/Core/Extensions/DiagnosticsExtensions.cs
@@ -486,8 +486,7 @@ public static class DiagnosticsExtensions
         if (provider?.TryGetGlobalDiagnosticValue(descriptor.Id, cancellationToken, out reportDiagnostic) == true)
             return IsEnabledReportDiagnostic(reportDiagnostic, descriptor);
 
-        if (descriptor.IsEnabledByDefault
-            && !string.IsNullOrEmpty(descriptor.Category))
+        if (!string.IsNullOrEmpty(descriptor.Category))
         {
             AnalyzerConfigOptions configOptions = analyzerOptions.AnalyzerConfigOptionsProvider.GetOptions(syntaxTree);
             string categoryKey = $"dotnet_analyzer_diagnostic.category-{descriptor.Category.ToLowerInvariant()}.severity";
--- a/src/Core/Extensions/DiagnosticsExtensions.cs
+++ b/src/Core/Extensions/DiagnosticsExtensions.cs
@@ -492,7 +492,7 @@ public static class DiagnosticsExtensions
             AnalyzerConfigOptions configOptions = analyzerOptions.AnalyzerConfigOptionsProvider.GetOptions(syntaxTree);
             string categoryKey = $"dotnet_analyzer_diagnostic.category-{descriptor.Category.ToLowerInvariant()}.severity";
 
-            if (configOptions.TryGetValue(categoryKey, out string severity)
+            if (configOptions.TryGetValue(categoryKey, out string? severity)
                 && TryParseReportDiagnostic(severity, out reportDiagnostic))
             {
                 return IsEnabledReportDiagnostic(reportDiagnostic, descriptor);
--- a/src/Core/Extensions/DiagnosticsExtensions.cs
+++ b/src/Core/Extensions/DiagnosticsExtensions.cs
@@ -437,6 +437,7 @@ public static class DiagnosticsExtensions
             descriptor,
             context.Symbol.Locations[0].SourceTree!,
             context.Compilation.Options,
+            context.Options,
             context.CancellationToken);
     }
 
@@ -446,6 +447,7 @@ public static class DiagnosticsExtensions
             descriptor,
             context.Node.SyntaxTree,
             context.Compilation.Options,
+            context.Options,
             context.CancellationToken);
     }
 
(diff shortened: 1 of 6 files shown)


src/Core/Extensions/DiagnosticsExtensions.cs
```diff
@@ -457,12 +459,54 @@ internal static bool IsEffective(
...
+
+            if (configOptions.TryGetValue(categoryKey, out string? severity)
+                && TryParseReportDiagnostic(severity, out reportDiagnostic))
+            {
+                return IsEnabledReportDiagnostic(reportDiagnostic, descriptor);
+            }
         }
 
+        if (compilationOptions.SpecificDiagnosticOptions.TryGetValue(descriptor.Id, out reportDiagnostic))
+            return IsEnabledReportDiagnostic(reportDiagnostic, descriptor);
```
josefpihrt: **Severity precedence looks inverted vs the other overload / `GetEffectiveSeverity`.**

This `AnalyzerOptions` overload resolves in the order:
1. tree diagnostic
2. global diagnostic
3. category severity
4. `SpecificDiagnosticOptions`
5. `IsEnabledByDefault`

The overload without `AnalyzerOptions` (and `GetEffectiveSeverity`) checks `SpecificDiagnosticOptions` *before* global / fallback. With the current order, e.g. category `none` would win over a command-line / ruleset entry that enables a specific diagnostic in `SpecificDiagnosticOptions`, and a category severity would also win over a per-id `Suppress` there.

Suggested order to align the overloads:
1. tree
2. `SpecificDiagnosticOptions`
3. global
4. category
5. `IsEnabledByDefault`

src/Core/Extensions/DiagnosticsExtensions.cs
```diff
@@ -457,12 +459,54 @@ internal static bool IsEffective(
...
+    {
+        SyntaxTreeOptionsProvider? provider = compilationOptions.SyntaxTreeOptionsProvider;
+
+        if (provider?.TryGetDiagnosticValue(syntaxTree, descriptor.Id, cancellationToken, out ReportDiagnostic reportDiagnostic) == true)
+            return IsEnabledReportDiagnostic(reportDiagnostic, descriptor);
+
+        if (provider?.TryGetGlobalDiagnosticValue(descriptor.Id, cancellationToken, out reportDiagnostic) == true)
+            return IsEnabledReportDiagnostic(reportDiagnostic, descriptor);
+
+        if (descriptor.IsEnabledByDefault
```
josefpihrt: **Category severity is only consulted when `IsEnabledByDefault` is true.**

Per the [.NET analyzer config docs](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-options), category severity is intended to bulk-configure rules in that category, including ones that are disabled by default. Gating on `IsEnabledByDefault` means an explicit `dotnet_analyzer_diagnostic.category-roslynator.severity = warning` would not treat a disabled-by-default rule as effective here.

For this PR’s goal (skip work when category is `none`) the gate is fine in practice for enabled-by-default Roslynator rules, but if this helper is the shared `IsEffective` path it may be more correct to always honor an explicit category severity when present (and let `IsEnabledReportDiagnostic` handle `default`).

src/Tests/Analyzers.Tests/RCS1047NonAsynchronousMethodNameShouldNotEndWithAsyncTests.cs
```diff
@@ -249,4 +249,30 @@ public IAsyncEnumerator<object> GetAsyncEnumerator()
...
+        await VerifyNoDiagnosticAsync(@"
+class C
+{
+    void FooAsync()
+    {
+    }
+}
+", options: Options.SetConfigOption("dotnet_analyzer_diagnostic.category-roslynator.severity", "none"));
+    }
 }
```
josefpihrt, line 278: Nice coverage for the #1751 repro (per-rule and category `none`). Consider a matching RCS1046 test as well, since the changelog and `IsAnyEffective` gate both cover RCS1046/RCS1047 together—otherwise a future regression could slip through on the async-suffix-required side only.

josefpihrt (commented):
Review notes for the AsyncSuffix / `IsEffective` severity fix. The direction looks right for #1751 (honor EditorConfig, including category-level `none`), with two concrete issues worth addressing before merge.

src/Core/Extensions/DiagnosticsExtensions.cs
josefpihrt: Cannot put `SpecificDiagnosticOptions` before category without breaking this fix in tests (and the #1751 EditorConfig scenario under the current verifier).

`CreateDocument` always calls `EnsureDiagnosticEnabled`, which writes `descriptor.DefaultSeverity` into `SpecificDiagnosticOptions`. Checking that before category would make `dotnet_analyzer_diagnostic.category-*.severity = none` a no-op whenever the test helper (or a compilation option that merely restates the default) is present.

Current order is therefore: tree per-id → global per-id → category → `SpecificDiagnosticOptions` → `IsEnabledByDefault`. Per-id EditorConfig still wins first. Aligning fully with `GetEffectiveSeverity` would need the verifier to stop injecting default severity into `SpecificDiagnosticOptions`.

src/Core/Extensions/DiagnosticsExtensions.cs
josefpihrt: Done. Category severity is now consulted even when `IsEnabledByDefault` is false, so an explicit `dotnet_analyzer_diagnostic.category-roslynator.severity` also applies to disabled-by-default rules such as RCS1046.

src/Tests/Analyzers.Tests/RCS1047NonAsynchronousMethodNameShouldNotEndWithAsyncTests.cs
josefpihrt: Added matching RCS1046 tests. Because `AsyncSuffixAnalyzer` uses `IsAnyEffective` for RCS1046 and RCS1047, the per-rule case also sets RCS1047 to `none` (RCS1046 is disabled by default, so RCS1047 would otherwise keep analysis running). The category `none` test covers both.

src/Core/Extensions/DiagnosticsExtensions.cs
josefpihrt: Done. `IsEffective` now uses the same order as `GetEffectiveSeverity`: tree → `SpecificDiagnosticOptions` → global → category → `IsEnabledByDefault`.

The verifier no longer calls `EnsureDiagnosticEnabled` when EditorConfig already sets `dotnet_diagnostic.{id}.severity` or `dotnet_analyzer_diagnostic.category-*.severity`, so category `none` is not overridden by the injected default.

PR #1798, opened 2026-08-09
fix(analyzers): ignore partial type parts for RCS1060

### LINKED ISSUE #1779: RCS1060 fires for partial classes in the same file
I have a pattern my company has used for many years where when we implement an interface we defined it all in a separate partial class in the same file.

```csharp
public partial class Foo
{
   // other members go in this partial class
}

public partial class Foo : IDisposable
{
   // only IDispoable members go in this partial class
   public void Dispose() {}
}
```

This pattern however triggers `RCS1060`, which I don't think it should.  It is the same type being declared.

josefpihrt:
## Summary
- Group type declarations by name within each namespace scope when counting types in a file.
- Skip additional `partial` declarations of an already-seen type so multiple partial parts of the same type do not trigger RCS1060.
- Add tests for the issue repro, partial + distinct type regression, and file-scoped namespace.

Fixes #1779

## Test plan
- [x] `dotnet test src/Tests/Analyzers.Tests --filter FullyQualifiedName~RCS1060`

### DIFF
--- a/src/Analyzers/CSharp/Analysis/DeclareEachTypeInSeparateFileAnalyzer.cs
+++ b/src/Analyzers/CSharp/Analysis/DeclareEachTypeInSeparateFileAnalyzer.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) .NET Foundation and Contributors. Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -46,9 +47,9 @@ public sealed class DeclareEachTypeInSeparateFileAnalyzer : BaseDiagnosticAnalyz
         MemberDeclarationSyntax firstTypeDeclaration = null;
         var isFirstReported = false;
 
-        Analyze(compilationUnitMembers);
+        Analyze(compilationUnitMembers, new HashSet<string>());
 
-        void Analyze(SyntaxList<MemberDeclarationSyntax> members)
+        void Analyze(SyntaxList<MemberDeclarationSyntax> members, HashSet<string> declaredTypeNames)
         {
             foreach (MemberDeclarationSyntax member in members)
             {
@@ -58,7 +59,7 @@ public sealed class DeclareEachTypeInSeparateFileAnalyzer : BaseDiagnosticAnalyz
                 if (member is NamespaceDeclarationSyntax namespaceDeclaration)
 #endif
                 {
-                    Analyze(namespaceDeclaration.Members);
+                    Analyze(namespaceDeclaration.Members, new HashSet<string>());
                 }
                 else if (SyntaxFacts.IsTypeDeclaration(member.Kind())
 #if ROSLYN_4_4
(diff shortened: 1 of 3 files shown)


src/Analyzers/CSharp/Analysis/DeclareEachTypeInSeparateFileAnalyzer.cs
```diff
@@ -58,14 +59,30 @@ void Analyze(SyntaxList<MemberDeclarationSyntax> members)
...
+                    string typeName = identifier.ValueText;
+
+                    if (declaredTypeNames.Contains(typeName)
+                        && member.Modifiers.Contains(SyntaxKind.PartialKeyword))
+                    {
+                        continue;
+                    }
+
+                    if (!declaredTypeNames.Contains(typeName))
+                        declaredTypeNames.Add(typeName);
```
josefpihrt: **Type identity should include arity, not just the identifier text.**

`Foo`, `Foo<T>`, and `Foo<T, U>` are distinct types that can legally coexist in the same namespace. Keying only on `identifier.ValueText` means this file would incorrectly skip the second/third declarations when they are `partial`:

```csharp
namespace N
{
    public partial class Foo { }
    public partial class Foo<T> { }
    public partial class Foo<T, U> { }
}
```

RCS1060 should still report here (three types in one file). Suggested key: name + type-parameter count (e.g. via `CSharpUtility.GetTypeParameterList(member)?.Parameters.Count ?? 0`), and a regression test for different arities.

src/Analyzers/CSharp/Analysis/DeclareEachTypeInSeparateFileAnalyzer.cs
```diff
@@ -58,14 +59,30 @@ void Analyze(SyntaxList<MemberDeclarationSyntax> members)
...
+
+                    if (identifier == default)
+                        continue;
+
+                    string typeName = identifier.ValueText;
+
+                    if (declaredTypeNames.Contains(typeName)
+                        && member.Modifiers.Contains(SyntaxKind.PartialKeyword))
+                    {
+                        continue;
```
josefpihrt: Nit: the contains/add sequence can be a single `Add` check with the same behavior:

```csharp
if (!declaredTypeNames.Add(typeKey)
    && member.Modifiers.Contains(SyntaxKind.PartialKeyword))
{
    continue;
}
```

(Only after the key includes arity, as in the other comment.)

src/Tests/Analyzers.Tests/RCS1060DeclareEachTypeInSeparateFileTests.cs
```diff
@@ -79,6 +79,66 @@ namespace N
...
+}
+
+public partial class Foo : IDisposable
+{
+    public void Dispose()
+    {
+    }
+}
 """);
     }
```
josefpihrt, line 258: Good coverage for the issue repro, partial + distinct type, and file-scoped namespace. Please also add the different-arity case called out on the analyzer so we do not regress false negatives for `Foo` / `Foo<T>` in one file.

josefpihrt (commented):
Review notes for the RCS1060 partial-type fix. The approach matches #1779 well (count distinct types, not every partial part), with one correctness gap around generic arity.

src/Analyzers/CSharp/Analysis/DeclareEachTypeInSeparateFileAnalyzer.cs
josefpihrt: Done. Type identity is now `(name, arity)` via `GetTypeParameterList` parameter count, so `Foo`, `Foo<T>`, and `Foo<T, U>` are counted as distinct types.

src/Analyzers/CSharp/Analysis/DeclareEachTypeInSeparateFileAnalyzer.cs
josefpihrt: Done. Collapsed the contains/add checks into `!declaredTypes.Add(typeKey)`.

src/Tests/Analyzers.Tests/RCS1060DeclareEachTypeInSeparateFileTests.cs
josefpihrt: Added `Test_PartialClassesWithDifferentArity_Reports` and a same-arity `Foo<T>` partial no-diagnostic case.

PR #1800, opened 2026-08-09
fix(analyzers): preserve space after is in RCS1172 code fix

### LINKED ISSUE #1704: Missing space after "is" in RCS1172 code fix
The code fix for RCS1172 has a missing space after the `is`:

`if (item as MyClass != null)` becomes `if (item isMyClass)`
`if (item as MyClass == null)` becomes `if (!(item isMyClass))`

josefpihrt:
## Summary

- Fix RCS1172 code fix producing `item isMyClass` / `!(item isMyClass)` by preserving whitespace between `is` and the type
- For `== null` checks, build `!(item is MyClass)` without stripping trivia via `WithoutTrivia().Parenthesize()`
- Add test for `item as MyClass == null`

Fixes #1704

## Test plan

- [x] `dotnet test src/Tests/Analyzers.Tests/Analyzers.Tests.csproj --filter "FullyQualifiedName~RCS1172"`

### DIFF
--- a/src/Analyzers.CodeFixes/CSharp/CodeFixes/UseIsOperatorInsteadOfAsOperatorCodeFixProvider.cs
+++ b/src/Analyzers.CodeFixes/CSharp/CodeFixes/UseIsOperatorInsteadOfAsOperatorCodeFixProvider.cs
@@ -86,7 +86,9 @@ public sealed class UseIsOperatorInsteadOfAsOperatorCodeFixProvider : BaseCodeFi
                     Token(SyntaxTriviaList.Empty, SyntaxKind.CloseParenToken, SyntaxTriviaList.Empty)));
         }
 
-        newNode = newNode.WithFormatterAnnotation();
+        newNode = newNode
+            .WithTriviaFrom(node)
+            .WithFormatterAnnotation();
 
         return document.ReplaceNodeAsync(node, newNode, cancellationToken);
     }
--- a/src/Analyzers.CodeFixes/CSharp/CodeFixes/UseIsOperatorInsteadOfAsOperatorCodeFixProvider.cs
+++ b/src/Analyzers.CodeFixes/CSharp/CodeFixes/UseIsOperatorInsteadOfAsOperatorCodeFixProvider.cs
@@ -2,6 +2,7 @@
 
 using System.Collections.Immutable;
 using System.Composition;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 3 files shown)


src/Analyzers.CodeFixes/CSharp/CodeFixes/UseIsOperatorInsteadOfAsOperatorCodeFixProvider.cs
```diff
@@ -57,17 +58,35 @@ private static Task<Document> RefactorAsync(
...
+            asExpressionInfo.Type.WithLeadingTrivia(typeLeadingTrivia));
 
         if (nullCheck.IsCheckingNull)
-            newNode = LogicalNotExpression(newNode.WithoutTrivia().Parenthesize()).WithTriviaFrom(newNode);
+        {
+            newNode = LogicalNotExpression(
+                ParenthesizedExpression(
+                    Token(SyntaxTriviaList.Empty, SyntaxKind.OpenParenToken, SyntaxTriviaList.Empty),
+                    newNode,
+                    Token(SyntaxTriviaList.Empty, SyntaxKind.CloseParenToken, SyntaxTriviaList.Empty)));
```
josefpihrt, line 86: The old null-check path used `WithTriviaFrom(newNode)` on the `LogicalNotExpression`, which kept leading/trailing trivia from the original comparison. The new `ParenthesizedExpression(...)` construction drops that.

Please restore trivia transfer onto the final expression (at least for the `== null` → `!(... is ...)` path), e.g. `.WithTriviaFrom(node)`, so comments on the replaced condition are not lost.

src/Analyzers.CodeFixes/CSharp/CodeFixes/UseIsOperatorInsteadOfAsOperatorCodeFixProvider.cs
```diff
@@ -57,17 +58,35 @@ private static Task<Document> RefactorAsync(
...
+        }
 
-        newNode = newNode
-            .Parenthesize()
-            .WithFormatterAnnotation();
+        newNode = newNode.WithFormatterAnnotation();
```
josefpihrt: Nit: the previous code always wrapped the result in `Parenthesize()` before the formatter annotation. Dropping that is probably fine for a full-node replace (and avoids the `WithoutTrivia()` footgun), but please confirm existing `!= null` → `is` tests still format as expected without the extra parentheses.

josefpihrt (commented):
Review notes for the RCS1172 `is` spacing fix. The whitespace preservation looks right for #1704; one trivia regression risk.

src/Analyzers.CodeFixes/CSharp/CodeFixes/UseIsOperatorInsteadOfAsOperatorCodeFixProvider.cs
josefpihrt: Done. The replacement now uses `WithTriviaFrom(node)` so trailing comments on the original comparison are preserved (`Test_EqualsNull_PreservesTrailingTrivia`).

src/Analyzers.CodeFixes/CSharp/CodeFixes/UseIsOperatorInsteadOfAsOperatorCodeFixProvider.cs
josefpihrt: Confirmed: the existing `!= null` → `item is MyClass` test still formats without extra parentheses, so leaving `Parenthesize()` off is correct for a full-node replace.

PR #1802, opened 2026-08-09
fix(analyzers): skip RCS1231 for sync Task-returning methods

### LINKED ISSUE #1702: RCS1231 is triggered for CancellationToken in non-async method
There is a discrepancy in the RCS1231 analyzer between sync and async methods

> RCS1231: Make parameter ref read-only
> https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1231/

``` csharp
internal sealed class Rcs1231
{
    // Async method => no warning
    public async Task DoThisWay(CancellationToken cancellationToken)
    {
        await Task.FromCanceled(cancellationToken).ConfigureAwait(false);
    }

    // Sync method returning Task => RCS1231 warning on CancellationToken parameter
    public Task DoThatWay(CancellationToken cancellationToken)
    {
        return Task.FromCanceled(cancellationToken);
    }
}
```

josefpihrt:
## Summary

- Align RCS1231 with the existing `async` exemption by skipping sync methods and local functions that return well-known task types (`Task`, `Task<T>`, `ValueTask`, `ValueTask<T>`)
- Fixes false positive on `CancellationToken` in sync methods like `return Task.FromCanceled(cancellationToken)`

Fixes #1702

## Test plan

- [x] `dotnet test src/Tests/Analyzers.Tests/Analyzers.Tests.csproj --filter "FullyQualifiedName~RCS1231"`

### DIFF
--- a/src/Analyzers/CSharp/Analysis/RefReadOnlyParameterAnalyzer.cs
+++ b/src/Analyzers/CSharp/Analysis/RefReadOnlyParameterAnalyzer.cs
@@ -59,6 +59,11 @@ public sealed class RefReadOnlyParameterAnalyzer : BaseDiagnosticAnalyzer
         if (methodDeclaration.Modifiers.ContainsAny(SyntaxKind.AsyncKeyword, SyntaxKind.OverrideKeyword))
             return;
 
+        IMethodSymbol methodSymbol = context.SemanticModel.GetDeclaredSymbol(methodDeclaration, context.CancellationToken);
+
+        if (methodSymbol?.ReturnType.IsWellKnownTaskType() == true)
+            return;
+
         Analyze(context, methodDeclaration, methodDeclaration.ParameterList, methodDeclaration.BodyOrExpressionBody());
     }
 
@@ -88,6 +93,11 @@ public sealed class RefReadOnlyParameterAnalyzer : BaseDiagnosticAnalyzer
         var localFunction = (LocalFunctionStatementSyntax)context.Node;
 
         if (localFunction.Modifiers.Contains(SyntaxKind.AsyncKeyword))
+            return;
+
+        IMethodSymbol methodSymbol = context.SemanticModel.GetDeclaredSymbol(localFunction, context.CancellationToken);
+
+        if (methodSymbol?.ReturnType.IsWellKnownTaskType() == true)
             return;
 
         Analyze(context, localFunction, localFunction.ParameterList, localFunction.BodyOrExpressionBody());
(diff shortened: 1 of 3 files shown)


src/Analyzers/CSharp/Analysis/RefReadOnlyParameterAnalyzer.cs
```diff
@@ -59,6 +59,11 @@ private static void AnalyzeMethodDeclaration(SyntaxNodeAnalysisContext context)
         if (methodDeclaration.Modifiers.ContainsAny(SyntaxKind.AsyncKeyword, SyntaxKind.OverrideKeyword))
             return;
 
+        IMethodSymbol methodSymbol = context.SemanticModel.GetDeclaredSymbol(methodDeclaration, context.CancellationToken);
+
+        if (methodSymbol?.ReturnType.IsWellKnownTaskType() == true)
+            return;
```
josefpihrt: This skips **all** parameters on any sync method/local function that returns `Task` / `Task<T>` / `ValueTask` / `ValueTask<T>`, not only `CancellationToken`.

That removes valid RCS1231 suggestions for large readonly structs, e.g.:

```csharp
Task<int> M(BigReadonlyStruct data) => Task.FromResult(data.Value);
```

If the false positive is specific to `CancellationToken` (or to tokens passed into `Task.FromCanceled` / similar), consider narrowing the exemption; if the intent is truly “never suggest `in` on sync task-returning methods,” please say so in the changelog/PR body.

src/Analyzers/CSharp/Analysis/RefReadOnlyParameterAnalyzer.cs
```diff
@@ -59,6 +59,11 @@ private static void AnalyzeMethodDeclaration(SyntaxNodeAnalysisContext context)
         if (methodDeclaration.Modifiers.ContainsAny(SyntaxKind.AsyncKeyword, SyntaxKind.OverrideKeyword))
             return;
 
+        IMethodSymbol methodSymbol = context.SemanticModel.GetDeclaredSymbol(methodDeclaration, context.CancellationToken);
```
josefpihrt: `Analyze(...)` already calls `GetDeclaredSymbol` on the same declaration a few lines later. Consider checking `ReturnType.IsWellKnownTaskType()` after that existing lookup (or passing the symbol in) to avoid the double semantic query.

josefpihrt (commented):
Review notes for skipping RCS1231 on sync Task-returning methods. Fixes the #1702 pattern, but the gate may be broader than intended. Also likely conflicts with #1807 on the same analyzer.

src/Analyzers/CSharp/Analysis/RefReadOnlyParameterAnalyzer.cs
josefpihrt: Narrowed the skip to `CancellationToken` parameters on sync methods that return a well-known task type. Other readonly structs still get RCS1231 (`Test_ReadOnlyStruct_SyncTaskReturningMethod`).

src/Analyzers/CSharp/Analysis/RefReadOnlyParameterAnalyzer.cs
josefpihrt: Done. The early `GetDeclaredSymbol` calls are gone; the task-type check now uses the symbol already obtained in `Analyze`.

PR #1804, opened 2026-08-09
fix(refactorings): offer RR0201 for more interpolated string spans

### LINKED ISSUE #1736: RR0201 Convert interpolated string to 'string.Format' not showing up
Refactory RR0201 option not showing up, while other refactoring and code fixes are working as expected in VS2022 and VS 2026.

josefpihrt:
## Summary

- Relax span checks for RR0201 (convert interpolated string to `string.Format`) and convert-to-concatenation so refactorings are offered when the caret is between interpolated string tokens, not only strictly inside the node span.
- Fix `IsRefactoringEnabled` to also resolve refactoring settings by RR id (e.g. `RR0201`), matching how Visual Studio options store them.
- Clarify in the `Roslynator.Analyzers` NuGet readme that refactorings require the separate `Roslynator.Refactorings` package (addresses the NuGet vs VS extension confusion in #1736).
- Add tests for RR0201 covering basic usage, cursor inside interpolation, on literal text, and verbatim interpolated strings.

Fixes #1736

## Test plan

- [x] `dotnet test src/Tests/Refactorings.Tests --filter FullyQualifiedName~RR0201`

### DIFF
--- a/src/Refactorings/CSharp/Refactorings/InterpolatedStringRefactoring.cs
+++ b/src/Refactorings/CSharp/Refactorings/InterpolatedStringRefactoring.cs
@@ -61,13 +61,13 @@ internal static class InterpolatedStringRefactoring
         }
 
         if (context.IsRefactoringEnabled(RefactoringDescriptors.ConvertInterpolatedStringToConcatenation)
-            && context.Span.IsEmptyAndContainedInSpanOrBetweenSpans(interpolatedString))
+            && context.Span.IsContainedInSpanOrBetweenSpans(interpolatedString))
         {
             ConvertInterpolatedStringToConcatenationRefactoring.ComputeRefactoring(context, interpolatedString);
         }
 
         if (context.IsRefactoringEnabled(RefactoringDescriptors.ConvertInterpolatedStringToStringFormat)
-            && context.Span.IsEmptyAndContainedInSpanOrBetweenSpans(interpolatedString))
+            && context.Span.IsContainedInSpanOrBetweenSpans(interpolatedString))
         {
             ConvertInterpolatedStringToStringFormatRefactoring.ComputeRefactoring(context, interpolatedString);
         }
--- a/src/Refactorings/CSharp/Refactorings/RefactoringContext.cs
+++ b/src/Refactorings/CSharp/Refactorings/RefactoringContext.cs
@@ -187,6 +187,9 @@ internal class RefactoringContext
         if (CodeAnalysisConfig.Instance.Refactorings.TryGetValue(refactoring.OptionKey, out enabled))
             return enabled;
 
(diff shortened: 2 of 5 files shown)


josefpihrt (commented):
Review notes for RR0201 span / option-key fixes. Looks good for #1736.

src/Refactorings/CSharp/Refactorings/RefactoringContext.cs
```diff
@@ -187,6 +187,9 @@ public bool IsRefactoringEnabled(RefactoringDescriptor refactoring)
         if (CodeAnalysisConfig.Instance.Refactorings.TryGetValue(refactoring.OptionKey, out enabled))
             return enabled;
 
+        if (CodeAnalysisConfig.Instance.Refactorings.TryGetValue(refactoring.Id, out enabled))
+            return enabled;
```
josefpihrt, line 191: Resolving by `refactoring.Id` (e.g. `RR0201`) after `OptionKey` is a nice catch for VS-style settings. Optional follow-up: mirror the same fallback anywhere else options are read by option-key only, so enable/disable stays consistent.

src/Refactorings/CSharp/Refactorings/InterpolatedStringRefactoring.cs
```diff
@@ -61,13 +61,13 @@ public static void ComputeRefactorings(RefactoringContext context, InterpolatedS
         }
 
         if (context.IsRefactoringEnabled(RefactoringDescriptors.ConvertInterpolatedStringToConcatenation)
-            && context.Span.IsEmptyAndContainedInSpanOrBetweenSpans(interpolatedString))
+            && context.Span.IsContainedInSpanOrBetweenSpans(interpolatedString))
```
josefpihrt, line 64: Relaxing `IsEmptyAndContainedInSpanOrBetweenSpans` → `IsContainedInSpanOrBetweenSpans` matches the reported caret-inside-interpolation / literal-text cases; the new tests cover the important positions well.

src/Refactorings/CSharp/Refactorings/RefactoringContext.cs
josefpihrt: `RefactoringContext.IsRefactoringEnabled` is the only call site that looked up `OptionKey`. `CodeAnalysisConfig.IsRefactoringEnabled` already keys by id. No other fallback to add in this PR.

src/Refactorings/CSharp/Refactorings/InterpolatedStringRefactoring.cs
josefpihrt: Agreed — no further change needed here.

PR #1805, opened 2026-08-09
fix(analyzers): report RCS1192 in generated-regex attribute args

### LINKED ISSUE #1714: RCS1192 - Not working for attribute string arguments
```cs
[GeneratedRegex(@"^[A-Za-z_][A-Za-z0-9_-]*$")] // RCS1192 expected here. 
private static partial Regex SchemaNameRegex();
```

josefpihrt:
## Summary

- Enable generated-code analysis for RCS1192 so verbatim string literals in attribute arguments are analyzed even when the containing partial method has a source-generated implementation (e.g. `[GeneratedRegex]`).
- Skip other generated code via `ShouldReport()` so only attribute-argument literals are reported in generated contexts.

Fixes #1714

## Test plan

- [x] `dotnet test src/Tests/Analyzers.Tests --filter FullyQualifiedName~RCS1192`


src/Analyzers/CSharp/Analysis/UnnecessaryUsageOfVerbatimStringLiteralAnalyzer.cs
```diff
@@ -29,7 +29,8 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
 
     public override void Initialize(AnalysisContext context)
     {
-        base.Initialize(context);
+        context.EnableConcurrentExecution();
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
```
josefpihrt, line 35: Skipping `base.Initialize` here is intentional so generated-code analysis can be enabled—good. Consider a one-line comment noting why `base.Initialize` is not called (otherwise it looks accidental next to other analyzers).

josefpihrt (commented):
Review notes for RCS1192 on generated-regex / partial generated implementations. Approach is sound for #1714; one scope question on generated code.

src/Analyzers/CSharp/Analysis/UnnecessaryUsageOfVerbatimStringLiteralAnalyzer.cs
```diff
@@ -110,12 +114,37 @@ private static void AnalyzeInterpolatedStringExpression(SyntaxNodeAnalysisContex
...
             context,
             DiagnosticRules.UnnecessaryUsageOfVerbatimStringLiteral,
             Location.Create(node.SyntaxTree, new TextSpan(start, 1)));
     }
 
+    private static bool ShouldReport(SyntaxNodeAnalysisContext context, SyntaxNode node)
+    {
+#if ROSLYN_4_0
+        if (!context.IsGeneratedCode)
+#else
```
josefpihrt: `ShouldReport` allows **any** `AttributeArgumentSyntax` once `IsGeneratedCode` is true. That fixes user partials tainted by a generated part, but it will also report verbatim strings inside attributes in fully generated trees.

If that is an accepted tradeoff, fine—please call it out. If not, tightening to “attribute on a partial method declaration” (or excluding generated *files*) would keep generator output quiet.

src/Analyzers/CSharp/Analysis/UnnecessaryUsageOfVerbatimStringLiteralAnalyzer.cs
josefpihrt: Added a comment explaining why `base.Initialize` is skipped (it would disable generated-code analysis).

src/Analyzers/CSharp/Analysis/UnnecessaryUsageOfVerbatimStringLiteralAnalyzer.cs
josefpihrt: Tightened `ShouldReport` so generated code is only flagged when the verbatim string is an attribute argument on a `partial` member (the GeneratedRegex / generated-implementation case). Other generated attribute arguments stay quiet.

PR #1807, opened 2026-08-09
fix(analyzers): skip RCS1231 for ref struct parameters

### LINKED ISSUE #1725: [RCS1231] On `ref struct`
This is my code
```c#
public readonly ref struct RefStruct;

public static class Methods
{
    public static void Main()
    {
        var item = new RefStruct();
        item = DoSomething(item);
    }

    public static RefStruct DoSomething(RefStruct @struct)
    {
        return @struct;
    }
}
```
And it'll give me an warning about `RCS1231: Make parameter ref read-only`. Yeah, I opened that warning for me.
But in this case, it shouldn't be added with `in`, or it'll give me another warning about `Cannot return local variable 'item' by reference because it is not a 'ref' variable`.

Is that possible to add an option about if it should work on `ref struct`? 

Thanks!

josefpihrt:
## Summary

- Skip RCS1231 (`Make parameter ref read-only`) for `ref struct` parameters; suggesting `in` can break returning the parameter by value.

Fixes #1725

## Test plan

- [x] `dotnet test src/Tests/Analyzers.Tests --filter FullyQualifiedName~RCS1231`

### DIFF
--- a/src/Analyzers/CSharp/Analysis/RefReadOnlyParameterAnalyzer.cs
+++ b/src/Analyzers/CSharp/Analysis/RefReadOnlyParameterAnalyzer.cs
@@ -142,6 +142,9 @@ public sealed class RefReadOnlyParameterAnalyzer : BaseDiagnosticAnalyzer
                 continue;
             }
 
+            if (type.IsRefLikeType)
+                continue;
+
             if (parameter.RefKind != RefKind.None)
                 continue;
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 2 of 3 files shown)


src/Analyzers/CSharp/Analysis/RefReadOnlyParameterAnalyzer.cs
```diff
@@ -142,6 +142,9 @@ private static void Analyze(
                 continue;
             }
 
+            if (type.IsRefLikeType)
+                continue;
```
josefpihrt, line 146: `IsRefLikeType` is the right gate—suggesting `in` can break returning/storing the parameter by value.

Note: this touches the same analyzer as #1802; expect a merge conflict when landing both, and please keep both exemptions.

josefpihrt (commented):
Review notes for skipping RCS1231 on `ref struct` parameters. Clean fix for #1725.

src/Analyzers/CSharp/Analysis/RefReadOnlyParameterAnalyzer.cs
josefpihrt: Agreed on `IsRefLikeType`. #1802 was narrowed to CancellationToken only, so the two exemptions should merge cleanly (this skip stays in the parameter loop).

PR #1809, opened 2026-08-09
fix(analyzers): skip RCS1102 for nested classes with instance members

### LINKED ISSUE #1474: RCS1102 False-positive when class is used as a type argument
**Product and Version Used**: 
Roslynator.Analyzers 4.12.3

**Steps to Reproduce**:
Have a class such as below:
```csharp
public class Create<TEntity, TOut> where TEntity : Common
```
And a usage of the class as seen here:
```csharp
_ = services.AddValidatorsFromAssemblyContaining<Create<Foo, Bar>>();
```

**Actual Behavior**:
RCS1102 is raised. If the fixer is used, a C# error is raised at the usage: 
```
'Create<Foo, Bar>': static types cannot be used as type arguments[CS0718]
```

**Expected Behavior**:
RCS1102 is not raised.

josefpihrt:
## Summary

- Skip RCS1102 when a nested class contains non-static members, since making the containing class `static` would require nested types to be static as well (breaking patterns like MediatR `Create<T>` with nested `Command`/`Handler` classes).
- Add regression test for generic type-argument usage (`Register<Create<Foo, Bar>>()`).

Fixes #1474

## Test plan

- [x] `dotnet test src/Tests/Analyzers.Tests --filter FullyQualifiedName~RCS1102`

### DIFF
--- a/src/Analyzers/CSharp/Analysis/MakeClassStaticAnalyzer.cs
+++ b/src/Analyzers/CSharp/Analysis/MakeClassStaticAnalyzer.cs
@@ -129,6 +129,12 @@ public sealed class MakeClassStaticAnalyzer : BaseDiagnosticAnalyzer
                             if (memberSymbol.DeclaredAccessibility.ContainsProtected())
                                 return false;
 
+                            if (namedTypeSymbol.TypeKind == TypeKind.Class
+                                && !CanBeNestedInStaticClass(namedTypeSymbol))
+                            {
+                                return false;
+                            }
+
                             break;
                         }
                         default:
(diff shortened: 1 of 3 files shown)


src/Analyzers/CSharp/Analysis/MakeClassStaticAnalyzer.cs
```diff
@@ -171,6 +177,49 @@ public static bool AnalyzeMembers(ImmutableArray<ISymbol> members)
...
+                    {
+                        return false;
+                    }
+
+                    break;
+                }
+            }
+        }
+
+        return true;
```
josefpihrt, line 220: Recursive check for nested classes with instance members looks correct for the reported pattern.

Residual (pre-existing, not introduced here): a *non-static* nested class that only has static members can still allow the outer class to be suggested `static`, which would then require the nested class to be marked `static` too. Only worth addressing if you want the analyzer/fix to be fully mechanical for nested types.

src/Tests/Analyzers.Tests/RCS1102MakeClassStaticTests.cs
```diff
@@ -174,6 +174,51 @@ public sealed class Class2 : Class1<T>
...
+    public static void Register<T>()
+    {
+    }
+
+    public static void M()
+    {
+        Register<Create<Foo, Bar>>();
+    }
+}
 ");
```
josefpihrt, line 222: Solid repro with nested `Command`/`Handler` plus `Register<Create<...>>()`. Thanks.

josefpihrt (commented):
Review notes for RCS1102 nested instance-member skip. Fixes the MediatR-style #1474 false positive.

src/Analyzers/CSharp/Analysis/MakeClassStaticAnalyzer.cs
josefpihrt: Leaving the nested static-only class case alone: it is pre-existing and not required to fix #1474. A follow-up can teach the analyzer/fix to mark nested types `static` as well if we want that to be fully mechanical.

src/Tests/Analyzers.Tests/RCS1102MakeClassStaticTests.cs
josefpihrt: Thanks — no further test change needed.

PR #1810, opened 2026-08-09
[5.0] Let consumers choose the Roslyn version used by the testing framework

josefpihrt:
## Problem

The Roslynator testing framework forced every test run onto a fixed Roslyn version (`4.14.0`), because `Roslynator.Testing.*` declared `Microsoft.CodeAnalysis.* >= 4.14.0`. Since NuGet dependency versions are minimums, a consumer whose analyzer targets an older Roslyn (e.g. the `roslyn3.8` / `roslyn4.7` buckets Roslynator itself ships) could not test against that version — NuGet unified the graph up to `4.14.0`.

## Change

- **Lower the testing packages' Roslyn floor to `3.8.0`** so the effective version is determined by the consumer's own `Microsoft.CodeAnalysis.*` reference. Verified the `Testing.*` projects use no API newer than 3.8. `build_core_and_testing` compiles and packs those projects against 3.8.
- **Decouple `Roslynator.Testing.Common` from the `Roslynator.Core` package** (its floor would otherwise leak back through). The ~6 internal helpers used (`StringBuilderCache`, `Hash`, the diagnostic comparers, `ToReportDiagnostic`) are duplicated into `Testing.Common` as `internal` types. `Roslynator.Core`'s own published floor is unchanged.
- **Add `RoslynatorTestRoslynVersion`** (default `5.0.0`, matching this repo's analyzer `Microsoft.CodeAnalysis.*` floor after #1787) so the in-repo suite pins the Roslyn version used at test *runtime*.
  - This property only pins the test projects' `Microsoft.CodeAnalysis.*` PackageReference. It does **not** switch `RoslynatorRoslynVersion` / analyzer compilation.
  - This repo cannot pin the suite below 5.0: test projects `ProjectReference` analyzers that require `Microsoft.CodeAnalysis.* >= 5.0.0`.
  - CI does not *run* the analyzer suite at `3.8.0`: that is the testing *package* compile floor for consumers, not a version this repo's analyzers can execute against.
- **Guard version-sensitive tests** (`SyntaxKindTests` and tests using C# 12/13 features or newer-Roslyn formatter output) with `ROSLYN_TEST_*` defines in `src/Tests/Directory.Build.props`.
- **Also fix** the roslyn4.7 `DefineConstants` list: `ROSLYN_4_2,ROSLYN_4_4` used a comma, so `ROSLYN_4_4` was never defined for that bucket. It is now semicolon-separated (already on `main` via #1787).

### Consumer action required

A test project that relied on the implicit `4.14.0` should now add its own reference:

```xml
<PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.14.0" />
```

The chosen version raises the *maximum* C# the parser can accept; test sources still parse at `CSharpParseOptions.Default` (not `Latest`) unless you set parse options / `LanguageVersion`. The same note is in the Testing.CSharp.Xunit / MSTest NuGet readmes.

A test project that used `Roslynator.Core` types via the old transitive dependency must add an explicit `Roslynator.Core` package reference.

## Verification

- Full suite green at `RoslynatorTestRoslynVersion=5.0.0` (default).
- Packed nuspecs confirm `Testing.Common`/`Testing.CSharp` floor at `3.8.0` with no `Roslynator.Core` dependency; `Roslynator.Core` still floors at the product Roslyn version.


src/Directory.Build.props
```diff
@@ -24,15 +24,21 @@
       <PropertyGroup>
         <RoslynatorRoslynVersion>4.7.0</RoslynatorRoslynVersion>
         <RoslynatorRoslynAnalyzersVersion>3.3.4</RoslynatorRoslynAnalyzersVersion>
-        <DefineConstants>$(DefineConstants);ROSLYN_3_8;ROSLYN_4_0;ROSLYN_4_2,ROSLYN_4_4;ROSLYN_4_7;ROSLYN_4_7_ONLY</DefineConstants>
+        <DefineConstants>$(DefineConstants);ROSLYN_3_8;ROSLYN_4_0;ROSLYN_4_2;ROSLYN_4_4;ROSLYN_4_7;ROSLYN_4_7_ONLY</DefineConstants>
```
josefpihrt, line 27: Nice catch on the side: `ROSLYN_4_2,ROSLYN_4_4` → `ROSLYN_4_2;ROSLYN_4_4` means the roslyn4.7 bucket actually defines `ROSLYN_4_4` now. Worth calling out explicitly in the PR body so reviewers do not treat it as an unrelated drive-by.

.github/workflows/build.yml
```diff
@@ -75,8 +75,33 @@ jobs:
...
+  # so the analyzers are exercised at the Roslyn version they are compiled against
+  # (the shipped roslyn4.7 bucket), not only at the newest Roslyn.
+  test_roslyn_versions:
+    if: github.ref_type != 'tag' || startsWith(github.ref_name, 'v')
+    needs: pre_build
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        roslyn: [ "4.7.0" ]
```
josefpihrt: Matrix only runs `4.7.0`, while the shipped testing packages now floor at `3.8.0`. That is probably enough for *this repo’s* suite, but it does not exercise the declared 3.8 API surface of `Testing.*`.

Also, the job comment says analyzers are exercised at the roslyn4.7 bucket, but `/p:RoslynatorTestRoslynVersion=` only pins the test projects’ `Microsoft.CodeAnalysis.*` PackageReference—it does not switch `RoslynatorRoslynVersion` / analyzer compilation. Please confirm (and maybe document) that restore/build still loads a coherent Roslyn graph when tests are on 4.7 and analyzers remain on the solution default.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- [Testing Framework] Lower Roslyn dependency of testing packages to 3.8.0 so that the Roslyn version is determined by the consumer's own `Microsoft.CodeAnalysis.*` reference instead of being forced to a fixed version ([PR](https://github.com/dotnet/roslynator/pull/1810))
```
josefpihrt, line 16: Consumer action required is documented clearly here—good. Optional: mirror the same note on the docs site / testing how-to, since this breaks anyone who relied on the implicit 4.14 pull.

josefpihrt (commented):
Review notes for making the testing framework’s Roslyn version consumer-selectable. Direction and changelog/breaking-change notes look strong; a few verification/CI clarifications.

src/Directory.Build.props
josefpihrt: Called out in the PR body, and added a comment on the `DefineConstants` line: the comma was a real bug (`ROSLYN_4_4` never defined for the 4.7 bucket).

.github/workflows/build.yml
josefpihrt: Corrected the CI job comment: `RoslynatorTestRoslynVersion` only pins test-project `Microsoft.CodeAnalysis.*` references; it does not switch analyzer compilation / `RoslynatorRoslynVersion`. Not adding a 3.8 matrix cell — that floor is for consumers of `Testing.*`, not a version this repo's analyzer suite is meant to execute against.

CHANGELOG.md
josefpihrt: The Testing.CSharp.Xunit / MSTest NuGet readmes already document the required `Microsoft.CodeAnalysis.CSharp.Workspaces` reference. Site docs live in josefpihrt.github.io and are out of scope for this PR.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - `Roslynator.Testing.Common`, `Roslynator.Testing.CSharp`, `Roslynator.Testing.CSharp.Xunit` and `Roslynator.Testing.CSharp.MSTest` now depend on `Microsoft.CodeAnalysis.*` `>= 3.8.0` (previously `>= 4.14.0`).
+  - `Roslynator.Testing.Common` no longer depends on `Roslynator.Core`.
+  - **Action required:** a test project that previously relied on the testing framework to pull in Roslyn 4.14.0 should now add its own reference, e.g. `<PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.14.0" />`. The chosen version also caps the C# language version available in test sources.
```
josefpihrt: **This is a breaking restore change, not only a ‘Changed’ behavior tweak.**

Consumers who depended on `Roslynator.Testing.*` to pull `Microsoft.CodeAnalysis.* >= 4.14.0` (and transitively `Roslynator.Core`) will silently restore 3.8.0 and lose Core unless they add the new `PackageReference`. The action-required bullet is clear, but Keep a Changelog / the repo’s SemVer claim would match better as a `### Breaking` subsection (or ship in a major alongside #1787).

Also spell out the Core break: any test project that used `Roslynator.Core` types without referencing that package now needs an explicit `Roslynator.Core` reference.

src/Tests/Analyzers.Tests/Analyzers.Tests.csproj
```diff
@@ -10,6 +10,14 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <!-- Some tests use C# features or rely on formatter behavior available only in newer Roslyn versions.
+       These symbols let such tests be excluded when the suite is pinned to an older RoslynatorTestRoslynVersion. -->
+  <PropertyGroup>
+    <DefineConstants Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(RoslynatorTestRoslynVersion)', '4.8'))">$(DefineConstants);ROSLYN_TEST_4_8_OR_GREATER</DefineConstants>
+    <DefineConstants Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(RoslynatorTestRoslynVersion)', '4.12'))">$(DefineConstants);ROSLYN_TEST_4_12_OR_GREATER</DefineConstants>
+    <DefineConstants Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(RoslynatorTestRoslynVersion)', '4.14'))">$(DefineConstants);ROSLYN_TEST_4_14_OR_GREATER</DefineConstants>
```
josefpihrt: **`ROSLYN_TEST_*` defines are split across two csprojs and missing from the others that also pin `RoslynatorTestRoslynVersion`.**

`Analyzers.Tests` defines 4.8 / 4.12 / 4.14; `CSharp.Tests` defines 4.11 / 4.12 / 4.14; `CodeFixes.Tests`, `Refactorings.Tests`, `Formatting.Analyzers.Tests`, and `CSharp.Workspaces.Tests` pin the package but cannot use the same `#if` guards.

CI at 4.7 happens to pass today, but the next C# 13 / formatter-sensitive test in those projects will fail the matrix with no symbol to ifdef. Prefer a `src/Tests/Directory.Build.props` that defines the full ladder (`4.8`, `4.11`, `4.12`, `4.14`) for every `*.Tests` project.

src/Tests/Testing.CSharp.Xunit/docs/NuGetReadme.md
```diff
@@ -2,6 +2,16 @@
...
 
+## Choosing the Roslyn version
+
+This package depends on `Microsoft.CodeAnalysis.*` with a low minimum version, so it does not force a particular Roslyn version on your test project. Add a reference to the Roslyn version you want your tests to run against - typically the same version your analyzer library is built for:
+
+```xml
+<PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.14.0" />
+```
+
+The chosen version also determines the maximum C# language version available in your test source code (for example, Roslyn 3.8 supports up to C# 9).
```
josefpihrt: **Package version is not the same as `LanguageVersion` used to parse test sources.**

`CSharpTestOptions` defaults to `CSharpParseOptions.Default`, which is the library’s *Default* language version, not `Latest`. Adding `Microsoft.CodeAnalysis.CSharp.Workspaces` 4.14.0 raises the *maximum* C# the parser can accept, but test strings still parse at Default (often a version behind Latest — e.g. no C# 13 `params ReadOnlySpan<T>` until the consumer calls `WithParseOptions` / `LanguageVersion.Latest`).

This repo’s own suite works around that with `#if ROSLYN_TEST_*` rather than bumping parse options. Worth one sentence here (and the MSTest readme) so consumers do not assume the `PackageReference` alone enables every C# version that Roslyn build supports.

josefpihrt (commented):
Detailed pass on the testing-framework Roslyn floor change. The approach is right: compile `Testing.*` against 3.8, drop the `Roslynator.Core` leak, pin this repo with `RoslynatorTestRoslynVersion`, and document the consumer `PackageReference`. CI is green at default 4.14 and at `test_roslyn_versions` 4.7.0; the earlier review nits on the 4.4 `DefineConstants` comma and the CI comment are addressed.

A few leftover product/docs issues before merge.

CHANGELOG.md
josefpihrt: Addressed: moved the Testing Framework entry under `### Breaking`, spelled out the explicit `Roslynator.Core` PackageReference requirement, and clarified that the chosen Roslyn package raises the *maximum* language version while parse options still default to `CSharpParseOptions.Default`.

src/Tests/Analyzers.Tests/Analyzers.Tests.csproj
josefpihrt: Addressed: added `src/Tests/Directory.Build.props` with the full `ROSLYN_TEST_*` ladder (`4.8` / `4.11` / `4.12` / `4.14`) for every project under `src/Tests`, and removed the duplicated PropertyGroups from `Analyzers.Tests` and `CSharp.Tests`.

src/Tests/Testing.CSharp.Xunit/docs/NuGetReadme.md
josefpihrt: Addressed in both the Xunit and MSTest NuGet readmes: the PackageReference raises the maximum C# the parser can accept, but test sources still parse at `CSharpParseOptions.Default` unless the consumer sets parse options / `LanguageVersion` (e.g. `Latest`).

